PR #70, opened 2026-01-27
fix: Check power status DB attribute when filtering Instances by status

nvlitagaki:
- Currently Instance status attribute in API layer presents both status and power status. However when filtering at DB layer we only check against DB status attribute
- This change checks both status and power status DB attribute to return a better set of matches aligning with API values

### DIFF
--- a/api/pkg/api/handler/instance_test.go
+++ b/api/pkg/api/handler/instance_test.go
@@ -421,7 +421,7 @@ func testInstanceBuildMachineInterface(t *testing.T, dbSession *cdb.Session, sub
 	return mi
 }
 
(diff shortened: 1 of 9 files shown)


thossain-nv (approved):
Looks good, thanks @nvlitagaki

hwadekar-nv (approved):

mxh-0xbb:
@nvlitagaki this PR has been stalled for a while.  Saw your comment about it not being a priority at present, but would like to either commit or abandon it.

thossain-nv:
Closing this for now, we can re-open when needed.

PR #102, opened 2026-02-10
feat: Add labels support in InstanceType

hwadekar-nv:
- Added 'labels' in API model
- Added 'labels' in DB model
- Unit test

It has dependancies (required)  >= https://github.com/NVIDIA/bare-metal-manager-core/tree/v0.1.0-rc1  

- bare-metal-manager-core InstanceType Labels support :- 
`-` api-model :-  https://github.com/NVIDIA/bare-metal-manager-core/blob/main/crates/api-model/src/instance_type/mod.rs#L275
`-` api-db :- https://github.com/NVIDIA/bare-metal-manager-core/blob/main/crates/api-db/src/instance_type.rs#L49


thossain-nv (commented):
Looks good @hwadekar-nv Can we fix the Git conflicts?

api/pkg/api/handler/instancetype.go
```diff
@@ -202,6 +202,7 @@ func (cith CreateInstanceTypeHandler) Handle(c echo.Context) error {
 		ControllerMachineType:    apiRequest.ControllerMachineType,
 		InfrastructureProviderID: ip.ID,
 		SiteID:                   &site.ID,
+		Labels:                   apiRequest.Labels,
```
thossain-nv: When creating, set this value to `{}` instead of `nil` when `apiRequest.Labels` is `nil`

PR #110, opened 2026-02-12
fix: Revise DPU Extension Service deletion and version deletion logic

thossain-nv:
- Currently when DPU Extension Services are deleted, REST layer relies on inventory update to remove the cache DB entry. Since deletion on Site is instantaneous, there is no need to wait for inventory processing
- When the last remaining version of a DPU Extension Service is deleted, the parent service is also deleted from Site. Updated version deletion handler to accommodate this
- When DPU Extension Service has multiple versions:
  - Deleting older version now immediately updates active versions list instead of waiting for inventory update
  - Deleting latest version now fetches most recent older version and re-populates version info

### DIFF
--- a/api/pkg/api/handler/dpuextensionservice.go
+++ b/api/pkg/api/handler/dpuextensionservice.go
@@ -1466,14 +1466,14 @@ func (ddesvh DeleteDpuExtensionServiceVersionHandler) Handle(c echo.Context) err
 	if len(remainingVersions) == 0 {
 		logger.Info().Msg("since deleted version was the last version, deleting DPU Extension Service from DB")
 		// Delete the DPU Extension Service record from DB
-		err := desDAO.Delete(ctx, nil, dpuExtensionService.ID)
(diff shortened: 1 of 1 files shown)


api/pkg/api/handler/dpuextensionservice.go
```diff
@@ -1437,6 +1444,66 @@ func (ddesvh DeleteDpuExtensionServiceVersionHandler) Handle(c echo.Context) err
...
+		}
+	}
+
+	fetchLatestRemainingVersion := false
+
+	// Check if this was the last version
+	if len(remainingVersions) == 0 {
+		logger.Info().Msg("since deleted version was the last version, deleting DPU Extension Service from DB")
+		// Delete the DPU Extension Service record from DB
+		err := desDAO.Delete(ctx, nil, dpuExtensionService.ID)
```
hwadekar-nv: I think here, do we require to pass `tx`?

kfelternv (approved):

PR #119, opened 2026-02-13
feat: Relocate RLA code to bare-metal-manager-rest

ddejong-spec:
This transfers the code from the internal RLA repo to being part of bare-metal-manager-rest.  Most of this is existing, already reviewed code.  The things that are not are:
- Anything outside of the rla directory
- Modules renamed to github location and copyright headers added
- Duplicate config that broke running in a container replaced with standard config module usage (rla/cmd/serve.go)
- Updates to rla/Makefile and rla/Docker*


zhaozhongn:
LGTM. We should get this in ASAP.

.github/workflows/build-push-docker.yml
```diff
@@ -195,6 +195,25 @@ jobs:
       release_tag: ${{ inputs.release_tag }}
     secrets: inherit
 
+  # Build and push carbide-rla
```
kunzhao-nv: We may also want to add rla int test in .github/workflows/lint-and-test.yml

thossain-nv (commented):
Overall looks good. We need the lint and test passing in Github workflow.

PR #122, opened 2026-02-13
feat: Add rack validate REST API endpoints

kunzhao-nv:
Add rack validation endpoints via RLA: 
- `GET /v2/org/{org}/forge/rack/{uuid}/validate` — validate single rack
- `GET /v2/org/{org}/forge/rack/{uuid}/validate?<filter>` — validate filtered racks

With activity, workflow, handler tests and OpenAPI spec.


api/pkg/api/handler/rack.go
```diff
@@ -349,9 +353,24 @@ func (garh GetAllRackHandler) Handle(c echo.Context) error {
 		OrderBy:        orderBy,
 	}
 
+	// Build a deterministic workflow ID from query params for dedup
+	// Sort query params to ensure consistent hash regardless of parameter order
+	sortedParams := make([]string, 0, len(c.QueryParams()))
```
kunzhao-nv: TODO: might need to extract out to a common func later

api/pkg/api/handler/rack.go
```diff
@@ -406,3 +428,312 @@ func (garh GetAllRackHandler) Handle(c echo.Context) error {
...
+		WorkflowExecutionTimeout: common.WorkflowExecutionTimeout,
+		TaskQueue:                queue.SiteTaskQueue,
+	}
+
+	ctx, cancel := context.WithTimeout(ctx, common.WorkflowContextTimeout)
+	defer cancel()
+
+	we, err := stc.ExecuteWorkflow(ctx, workflowOptions, "ValidateComponents", rlaRequest)
+	if err != nil {
+		logger.Error().Err(err).Msg("failed to execute ValidateComponents workflow")
```
thossain-nv: We're missing the part where we detect the timeout and cancel the scheduled workflow since we're erroring the HTTP request.

site-workflow/pkg/workflow/rack.go
```diff
@@ -102,3 +102,40 @@ func GetRacks(ctx workflow.Context, request *rlav1.GetListOfRacksRequest) (*rlav
 
 	return &response, nil
 }
+
+// ValidateComponents is a workflow to validate components by comparing expected vs actual state via RLA.
+// Supports validating a single rack, multiple racks with filters, or all racks in a site.
+func ValidateComponents(ctx workflow.Context, request *rlav1.ValidateComponentsRequest) (*rlav1.ValidateComponentsResponse, error) {
```
thossain-nv: Same as above, should be `ValidateRackComponents`

workflow-schema/rla/proto/v1/rla.proto
```diff
@@ -1,18 +1,3 @@
-// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
thossain-nv: We should keep the copyright notice.

PR #134, opened 2026-02-18
fix: Set NVLinkPartition status based on response from Site

hwadekar-nv:
- CreateNvlinkLogicalPartition workflow and activity now return the NVLinkLogicalPartition object as their result.
- The API handler uses this result to persist the correct status (e.g., ready) to the DB, rather than relying on the initial creation state.


hwadekar-nv:
@thossain-nv Based on example given by Temporal Go SDK https://github.com/temporalio/samples-go/blob/main/helloworld/worker/main.go#L23 even workflow signature changed where it defined, no explicit return type definition is needed. Temporal infers it from the function signature. The current registration we have is correct

api/pkg/api/handler/nvlinklogicalpartition.go
```diff
@@ -348,8 +353,25 @@ func (cibph CreateNVLinkLogicalPartitionHandler) Handle(c echo.Context) error {
 	// set committed so, deferred cleanup functions will do nothing
 	txCommitted = true
 
+	// update the db record for NVLink Logical Partition with the status from the workflow
+	// If we run into an error, we'll log it but won't return error
+	unvllp := nvllp
+	ussd := ssd // Use original status detail by default
```
thossain-nv: Create a variable here: `ssds := []cdbm.StatusDetail{*ssd}`

common/pkg/util/nvlinklogicalpartition.go
```diff
@@ -0,0 +1,72 @@
+/*
```
thossain-nv: Let's move these functions back to workflow as we don't want `common` gaining knowledge of DB.

thossain-nv (approved):
Thanks @hwadekar-nv, looks good, one last comment.

PR #137, opened 2026-02-18
feat: Add site-level stats endpoints for GPU, instance type, and tenant allocation

parmani-nv:


hwadekar-nv:
Thank you, @parmani-nv, for the PR. Recently, I have learned that whenever we are changing spec, we should run ‘make preview-openapi’ or ‘make publish-openapi’. This will help us verify if we are breaking the specification structure or missing anything.

api/pkg/api/handler/stats.go
```diff
@@ -0,0 +1,778 @@
...
+				UsedMachineStats: used,
+				MaxAllocatable:   maxAlloc,
+				Allocations:      apiAllocs,
+			})
+		}
+
+		tenantStatsMap[tID] = ts
+	}
+
+	result := lo.MapToSlice(tenantStatsMap, func(_ uuid.UUID, ts *model.APITenantInstanceTypeStats) model.APITenantInstanceTypeStats {
```
thossain-nv: We should use native Go libraries when we can https://pkg.go.dev/maps#Values

api/pkg/api/handler/stats.go
```diff
@@ -0,0 +1,729 @@
...
+}
+
+// buildInstanceTypeStats builds a single APIMachineInstanceTypeStats for one instance type
+func buildInstanceTypeStats(
+	it cdbm.InstanceType,
+	itMachines []cdbm.Machine,
+	itConstraints []cdbm.AllocationConstraint,
+	itUsed map[uuid.UUID]*model.APIUsedMachineStats,
+	tenantITUsed map[uuid.UUID]map[uuid.UUID]*model.APIUsedMachineStats,
+) model.APIMachineInstanceTypeStats {
```
thossain-nv: Let's move this to stats model file and call it `NewAPIMachineInstanceTypeStats`

api/pkg/api/handler/stats.go
```diff
@@ -0,0 +1,778 @@
...
+	}
+
+	siteIDStr := c.QueryParam("siteId")
+	if siteIDStr == "" {
+		return cerr.NewAPIErrorResponse(c, http.StatusBadRequest, "siteId query parameter is required", nil)
+	}
+	site, err := common.GetSiteFromIDString(ctx, nil, siteIDStr, h.dbSession)
+	if err != nil {
+		logger.Error().Err(err).Str("siteId", siteIDStr).Msg("error parsing or retrieving site")
+		return cerr.NewAPIErrorResponse(c, http.StatusBadRequest, "Invalid siteId", nil)
```
pbreton: Should we expand, somethin like Invalid Site ID provided in query param

PR #138, opened 2026-02-18
chore: Rename local deployment elements

pbreton:
- rename elektraserver to mock-core
- rename cluster carbide-local to carbide-rest-local


thossain-nv (approved):
Looks good, thanks @pbreton

PR #140, opened 2026-02-19
chore: Consolidate kustomize objects into carbide-rest namespace

pbreton:


thossain-nv (approved):
Thanks @pbreton, seems ready to merge.

PR #145, opened 2026-02-19
fix: Resolve data race in RLA workerpool metrics

kunzhao-nv:
Resolve data race in workerpool metrics by using consistent atomic synchronization.
Failure example: https://github.com/NVIDIA/bare-metal-manager-rest/actions/runs/22198333184/job/64204427306?pr=128


pbreton (approved):
~~Not sure how it got passed CI/CD in the first place but currently failing my branch rebased from current main...~~
It is flaky: got in on 'luck'.
```go test -v -race -count=1 ./rla/pkg/workerpool/...```

PR #161, opened 2026-02-25
feat: Add Powershelf Manager service as a component

ddejong-spec:
OSRB https://nvbugspro.nvidia.com/bug/5604404


spydaNVIDIA:
> @ddejong-spec Seems like we're adding Hashicorp libraries which are in general not friendly to OSS development. Is Vault access a requirement for PSM?

@thossain-nv yes, for now, it is a requirement. later on, we may deprecate this.

cc @zhaozhongn

PR #165, opened 2026-02-26
feat: Add bmmcli TUI interactive mode with config selector

kfelternv:
- bmmcli tui (alias: bmmcli i) starts an interactive REPL
- Config selector shown at startup when multiple config*.yaml files exist in ~/.bmm/ — solves multi-env (local/staging/prod) switching
- REPL features: inline autocomplete, tab completion, scope filtering (scope site / scope vpc), org switching, command history via up-arrow windowed selector, Ctrl+D to exit
- All 20+ resource types supported: site, vpc, subnet, instance, machine, operating-system, ip-block, allocation, rack, sku, etc.
- Machine list shows VPC column via instance cross-reference
- Instance list resolves VPC/site IDs to names
- Windowed select widget (12 items) handles large lists gracefully
- login command does real NGC API key exchange (not cached token)
- --config global flag and BMM_CONFIG env var for non-TUI commands
- ExchangeAPIKey exported for TUI log flow
- Pure Go stdlib + golang.org/x/term, no external TUI dependencies


cli/pkg/auth.go
```diff
@@ -111,6 +110,49 @@ func InitCommand() *cli.Command {
 	}
 }
 
+// ExchangeAPIKey exchanges an NGC API key for a bearer token, updates the config,
+// and saves it. Returns the new token.
+func ExchangeAPIKey(cfg *ConfigFile, configPath string) (string, error) {
+	if cfg.Auth.APIKey == nil || cfg.Auth.APIKey.Key == "" {
+		return "", fmt.Errorf("no NGC API key configured")
+	}
+	authnURL := cfg.Auth.APIKey.AuthnURL
```
thossain-nv: We should call `AuthnURL` - `TokenExchangeURL`

Issue #86, opened 2026-02-03
Allow explicit ID and VNI in VPC partition creation requests for REST API


bcavnvidia:

AlexGridnev:
@dbl-nv please verify

PR #166, opened 2026-02-26
feat: Added support for explicit VPC ID and VNI

bcavnvidia:
## Summary

Some API consumers are expected to request a specific VPC ID and VNI instead of relying only on auto-allocation, and they will need a way to track which VPCs have had their VNIs auto-allocated vs explicitly allocated.

This PR adds end-to-end support for explicit VPC ID and VNI assignment on VPC creation and separates requested VNI from active VNI in API/DB/workflow.

  - Existing clients are unaffected unless they opt into VNI on creation.
  - Response payload now includes additional VNI fields.

Related issue: https://github.com/NVIDIA/bare-metal-manager-rest/issues/86

## Validation
  - Unit tests were updated in changed packages to cover the new behavior.


db/pkg/db/model/vpc.go
```diff
@@ -100,6 +100,7 @@ type Vpc struct {
 	NVLinkLogicalPartition                 *NVLinkLogicalPartition                 `bun:"rel:belongs-to,join:nvlink_logical_partition_id=id"`
 	NetworkVirtualizationType              *string                                 `bun:"network_virtualization_type"`
 	ControllerVpcID                        *uuid.UUID                              `bun:"controller_vpc_id,type:uuid"`
+	ActiveVni                              *int                                    `bun:"active_vni,type:integer"`
```
thossain-nv: Shouldn't we just name them according to API e.g. `RequestedVni` and `Vni`?  We should add the `Vni` attribute up here btw.

workflow/pkg/activity/vpc/vpc.go
```diff
@@ -267,8 +267,14 @@ func (mv ManageVpc) UpdateVpcInDB(ctx context.Context, transactionID *cwssaws.Tr
 					networkVirtualizationType = cdb.GetStrPtr(cdbm.VpcEthernetVirtualizer)
 				}
 
+				var activeVni *int
```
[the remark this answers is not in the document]
bcavnvidia: Oh, been a while. :smile:  For some reason I thought we started calling from `UpdateVpcsInDB`.

api/pkg/api/handler/vpc.go
```diff
@@ -369,12 +372,22 @@ func (cvh CreateVPCHandler) Handle(c echo.Context) error {
 	if *vpc.NetworkVirtualizationType == cwssaws.VpcVirtualizationType_FNN.String() {
 		nwvt = cwssaws.VpcVirtualizationType_FNN
 	}
+
+	vni, err := wutil.GetIntPtrToUint32Ptr(apiRequest.Vni)
+	// We already validate the VNI value in the model validate call, so no err
+	// is ever expected by this point, but we still need to handle it.
+	if err != nil {
+		logger.Error().Err(err).Msg("failed to convert vni to uint32 pointer")
+		return cerr.NewAPIErrorResponse(c, http.StatusInternalServerError, "Failed to convert vni value", nil)
```
thossain-nv: Maybe a more descriptive message would be good? `VNI value conversion failed despite passed validation`

api/pkg/api/handler/vpc.go
```diff
@@ -311,6 +312,7 @@ func (cvh CreateVPCHandler) Handle(c echo.Context) error {
 
 	// Create VPC
 	vpcInput := cdbm.VpcCreateInput{
+		ID:                        apiRequest.ID,
```
thossain-nv: I think we should check if there is an existing VPC with ID. Sometimes scripts can send stuff repeatedly and it'd be better to return a proper response instead of `Failed to create VPC, DB error` which this would result in case of a dupe.

thossain-nv (commented):
@bcavnvidia Went through the PR in more details, left a few notes.

PR #180, opened 2026-03-02
feat: Allow filtering Machines by the presence of associated Instance

nvlitagaki:
- Provider and privileged Tenants can filter Machines to those that have an Instance associated and those that don't


api/pkg/api/handler/machine.go
```diff
@@ -474,6 +475,28 @@ func (gamh GetAllMachineHandler) Handle(c echo.Context) error {
...
 	}
 
+	//	Check if `hasInstance` query params
+	qHasInstance := c.QueryParam("hasInstance")
+	if qHasInstance != "" {
+		hi, serr := strconv.ParseBool(qHasInstance)
+		if serr != nil {
+			return cerr.NewAPIErrorResponse(c, http.StatusBadRequest, "Invalid value specified for `hasInstance` in query", nil)
+		}
+		if infrastructureProvider == nil && !tenant.Config.TargetedInstanceCreation {
```
thossain-nv: We don't need this check since a little bit above in the handler we ensure that only Providers and privileged Tenants can proceed.

PR #190, opened 2026-03-04
refactor: Refactor Firmware and BringUp workflows to rule-based execution

kunzhao-nv:
- Refactor FirmwareControl and BringUp workflows from hardcoded sequential logic to the same rule-based execution pattern already used by PowerControl
- Extract shared `executeRuleBasedOperation` into helpers.go, eliminating duplicated stage iteration logic across all three workflows
- Add new action executors (AllowBringUp, WaitBringUp, FirmwareControl start+poll) and update PowerControl action to support cross-workflow usage via ParamOperation

### DIFF
--- a/rla/internal/task/executor/temporalworkflow/workflow/injectexpectation_test.go
+++ b/rla/internal/task/executor/temporalworkflow/workflow/injectexpectation_test.go
@@ -95,9 +95,9 @@ func TestInjectExpectationWorkflow(t *testing.T) {
 		},
 		"activity failure returns error": {
 			components:    computeOnly,
-		activityError: errors.New("component manager service unavailable"),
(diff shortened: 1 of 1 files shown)


rla/internal/task/executor/temporalworkflow/workflow/actions.go
```diff
@@ -275,16 +376,86 @@ func verifyPowerStatus(
 	}
 }
 
-// verifyReachability polls GetPowerStatus for multiple component types
-// until all are reachable.
+// executeAllowBringUpAction opens the Carbide power-on gate for the target
```
jw-nvidia: The actions should be unaware of the underlying component services. Avoid mentioning Carbide specific information.

PR #192, opened 2026-03-04
feat: Port NV-Switch manager into the Bare-Metal-Manager-Rest repo

spydaNVIDIA:


docker/local/Dockerfile.carbide-nsm
```diff
@@ -0,0 +1,53 @@
...
+
+WORKDIR /workspace/nvswitch-manager
+
+# Build with version information injected via ldflags
+RUN go build -ldflags "-extldflags '-static'" \
+    -o /app/nsm \
+    .
+
+# Runtime stage — NSM requires shell tools for firmware update scripts
+FROM alpine:3.19 AS final
```
ddejong-spec: This should be changed to distroless instead of alpine.  Compare the old dockerfiles from the RLA and PSM gates to the new ones for the tweaks you'll need to make it work.

nvswitch-manager/assets/ytl/nvl8-gp1-switches.json
```diff
@@ -0,0 +1,58 @@
+{
```
ddejong-spec: This is internal NVIDIA configuration and should not be present!

.gitattributes
```diff
@@ -0,0 +1,4 @@
+# Firmware binaries
```
kunzhao-nv: Do we need LFS here?

nvswitch-manager/cmd/serve.go
```diff
@@ -0,0 +1,182 @@
...
+			FirmwareConf: svc.FirmwareConfig{
+				PackagesDir:       firmwarePackagesDir,
+				FirmwareDir:       firmwareFirmwareDir,
+				NumWorkers:        firmwareNumWorkers,
+				SchedulerInterval: time.Duration(firmwarePollSeconds) * time.Second,
+			},
+		},
+	)
+
+	log.Printf("New service is created with port: %+v, data store type: %s, vault address: %s, vault token: %s", port, datastoreType, vaultAddress, vaultToken)
```
kunzhao-nv: The log order and content should be tuned, otherwise there may be case, we see "New service is created" first and then in L168, we see "failed to create"

spydaNVIDIA:
> The vulnerability scan was reporting 60 issues before, with these changes it is now reporting 63. It was a huge mistake in my opinion doing the initial release with anything more than zero, making it easy to overlook stuff like that, but we should at least not let the number go up any more.

@ddejong-spec I fixed the vulnerabilities assigned. The above comment is not being updated because of a bug in our CI/CD workflow. Larry has a fix open: https://github.com/NVIDIA/bare-metal-manager-rest/pull/210

Now, this MR introduces no new vulnerabilities.

nvswitch-manager/pkg/db/config.go
```diff
@@ -0,0 +1,128 @@
...
+		return errors.New("valid credential is required")
+	}
+
+	return nil
+}
+
+// BuildDSN builds the Data Source Name (DSN) string for connecting to
+// the database.
+func (c *Config) BuildDSN() string {
+	dsn := fmt.Sprintf(
```
kunzhao-nv: Should we do url encode on dsn, otherwise it might not handle special char like "@" in password?

nvswitch-manager/pkg/firmwaremanager/strategy_ssh.go
```diff
@@ -0,0 +1,593 @@
...
+	targetAddr := fmt.Sprintf("%s@%s:%s",
+		tray.NVOS.Credential.User,
+		tray.NVOS.IP.String(),
+		remotePath,
+	)
+
+	// Get the port from NVOS config (uses custom port if set, otherwise default 22)
+	port := fmt.Sprintf("%d", tray.NVOS.GetPort())
+
+	cmd := exec.Command("sshpass", "-p", tray.NVOS.Credential.Password.Value,
```
kunzhao-nv:
```
"sshpass", "-p"
```
is not recommended. It is insecure because the password appears in plaintext and process listings. Better use sshpass -e (env var SSHPASS) or sshpass -f (file descriptor) instead of -p.

kunzhao-nv (approved):

PR #194, opened 2026-03-04
ci: Update style check workflow to error on `go fmt` failure

thossain-nv:
Currently the workflow is silently checking file format but not raising an error.

### DIFF
--- a/.github/workflows/slack-notifications.yml
+++ b/.github/workflows/slack-notifications.yml
@@ -25,8 +25,10 @@ jobs:
     steps:
       - name: Send Slack Notification
         uses: slackapi/slack-github-action@v2.1.1 # Use a pre-existing action
+        env:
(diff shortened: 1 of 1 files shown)


.github/workflows/lint-and-test.yml
```diff
@@ -44,14 +44,22 @@ jobs:
       - name: Download dependencies
         run: go mod download
 
-      - name: Install revive
-        run: go install github.com/mgechev/revive@v1.3.9
+      - name: Install and run revive
+        run: |
+          go install github.com/mgechev/revive@v1.3.9
```
hwadekar-nv: Do we need this in Dockerfile?

hwadekar-nv: Got it . I thought it required to bo configured in dockerfile too

PR #217, opened 2026-03-09
feat: Infer Tenant/Provider context from org when retrieving Allocation/Instance Type

parmani-nv:
## Description

Removes the need for callers to pass explicit `infrastructureProviderId` and `tenantId` query parameters on Allocation and InstanceType read endpoints by inferring the caller's provider/tenant identity directly from their org membership, aligning these handlers with the pattern already established by the IPBlock and Site handlers.

**Allocation (GetAll / GetByID)**
- Replaced the manual provider/tenant query-param validation block with `IsProviderOrTenant`.
- GetAll now uses the dual-query-merge pattern (mapset union): results from the provider perspective and the tenant perspective are merged, so dual-role users receive a combined view in a single call.
- `infrastructureProviderId` is fully deprecated. `tenantId` is demoted to an optional provider-side filter (ignored when the caller only has the Tenant role).

**InstanceType (GetAll / GetByID)**
- Migrated from `GetIsProviderRequest` (which required a disambiguation query param for dual-role users) to `IsProviderOrTenant` with the same dual-query-merge pattern.
- `infrastructureProviderId` and `tenantId` query params are deprecated.

**Instance / InstanceBatch Create**
- `tenantId` in the request body is now optional; when omitted, the tenant is inferred from the caller's org membership.
- `ValidateOrgMembership` calls removed (already performed inside `IsProviderOrTenant` / `GetTenantForOrg`).

**Swagger / OpenAPI fixes**
- Corrected `siteId` annotation on the Allocation and InstanceType endpoints from `required: true` to `required: false`, matching actual runtime behaviour (the field was never enforced as required in code).

**Tests**
- Updated allocation and instancetype test suites to exercise inferred-ID paths and remove now-invalid required-param scenarios.

---

## Type of Change

---

## Services Affected

---

## Related Issues (Optional)

---

## Breaking Changes
- [x] This PR contains breaking changes

**What changes:**
The `infrastructureProviderId` and `tenantId` query parameters on `GET /allocation` and `GET /instance-type` (list and by-ID) are now **deprecated and no longer required**. Clients that currently pass them will not break — provider-side `tenantId` still works as a filter — but the enforcement that *at least one must be present* is removed. Any integration test or client that asserted a 400 response when those params were absent will need to be updated.

The `tenantId` field in the `POST /instance` and `POST /instance/batch` request body is now optional. Existing callers that supply it continue to work; the value is cross-validated against the org's inferred tenant.


api/pkg/api/handler/allocation.go
```diff
@@ -752,13 +670,72 @@ func (gaah GetAllAllocationHandler) Handle(c echo.Context) error {
...
+			return cutil.NewAPIErrorResponse(c, http.StatusInternalServerError, "Failed to retrieve Allocations, DB error", nil)
+		}
+		for _, al := range providerAllocations {
+			mergedAllocationIDs.Add(al.ID)
+		}
+	}
+
+	if tenant != nil {
+		tenantFilter := sharedFilter
+		tenantFilter.TenantIDs = []uuid.UUID{tenant.ID}
```
thossain-nv: Theoretically Tenants could filter by Provider. But we can skip it for now since NICo being used in multi-Provider scenario is unlikely.

api/pkg/api/handler/allocation.go
```diff
@@ -854,15 +831,15 @@ func NewGetAllocationHandler(dbSession *cdb.Session, tc temporalClient.Client, c
 
 // Handle godoc
 // @Summary Retrieve the Allocation
-// @Description Retrieve the Allocation
+// @Description Retrieve the Allocation from both Provider and Tenant perspectives
```
thossain-nv: We don't need to say this for a specific Allocation.

api/pkg/api/handler/allocation.go
```diff
@@ -920,86 +870,25 @@ func (gah GetAllocationHandler) Handle(c echo.Context) error {
...
-			if !ok {
-				return cutil.NewAPIErrorResponse(c, http.StatusForbidden, "User does not have Tenant Admin role with org", nil)
-			}
-
-			isAssociated = true
+		if err == cdb.ErrDoesNotExist {
+			return cutil.NewAPIErrorResponse(c, http.StatusNotFound, "Could not find Allocation with specified ID", nil)
 		}
+		logger.Error().Err(err).Msg("error retrieving Allocation DB entity")
+		return cutil.NewAPIErrorResponse(c, http.StatusInternalServerError, "Could not retrieve Allocation entity", nil)
```
thossain-nv: Should be `Failed to retrieve Allocation, DB error`

api/pkg/api/handler/allocation.go
```diff
@@ -495,15 +496,15 @@ func NewGetAllAllocationHandler(dbSession *cdb.Session, tc temporalClient.Client
 
 // Handle godoc
 // @Summary Retrieve all Allocations
-// @Description Retrieve all Allocations that Provider has created for a particular Site
+// @Description Retrieve all Allocations from both Provider and Tenant perspectives for the org
```
thossain-nv: We can say `Retrieve all Allocations relevant to current org`

api/pkg/api/handler/instancetype.go
```diff
@@ -836,15 +839,15 @@ func NewGetInstanceTypeHandler(dbSession *cdb.Session, tc temporalClient.Client,
 
 // Handle godoc
 // @Summary Get details of an Instance Type
-// @Description Get details of a specific Instance Type by ID
+// @Description Get details of a specific Instance Type from both Provider and Tenant perspectives
```
thossain-nv: `Retrieve details of a specific Instance Type by ID`

thossain-nv (commented):
Overall looks good, a few changes needed. Also CI is failing.

api/pkg/api/handler/instancetype.go
```diff
@@ -623,95 +613,100 @@ func (gaith GetAllInstanceTypeHandler) Handle(c echo.Context) error {
...
-	// 		- Either single siteID passed to the query
-	//		- All sites associated with the TenantID
-	filter := cdbm.InstanceTypeFilterInput{InfrastructureProviderID: infrastructureProviderID, SiteIDs: siteIDs, Status: status, SearchQuery: searchQuery}
-	if excludeUnallocated {
-		if tenantID == nil {
-			return cutil.NewAPIErrorResponse(c, http.StatusForbidden, "Only Tenant can specify query param `excludeUnallocated`", nil)
+		if excludeUnallocated {
+			tenantFilter.TenantIDs = []uuid.UUID{tenant.ID}
+		}
+		tenantInstanceTypes, _, err := itDAO.GetAll(ctx, nil, tenantFilter, nil, nil, cdb.GetIntPtr(cdbp.TotalLimit), nil)
```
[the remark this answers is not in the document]
thossain-nv: @parmani-nv Seems like a valid point?

PR #257, opened 2026-03-13
feat: Add REST API models and endpoints for VPC peering

hanyux-nv:
## Description
Add API endpoints for VPC peering including
- `POST /v2/org/{org}/carbide/vpc-peering` create a VPC peering between two VPCs
- `GET /v2/org/{org}/carbide/vpc-peering` get all VPC peerings visible to caller within site
- `GET /v2/org/{org}/carbide/vpc-peering/{id}` get a VPC peering by ID
- `DELETE /v2/org/{org}/carbide/vpc-peering/{id}` delete a VPC peering by ID

## Type of Change

## Services Affected

## Breaking Changes

## Testing


api/pkg/api/handler/vpcpeering.go
```diff
@@ -0,0 +1,939 @@
...
+		if peering.Vpc1ID == vpc2ID || peering.Vpc2ID == vpc2ID {
+			return cutil.NewAPIErrorResponse(c, http.StatusConflict, "VPC peering already exists between these VPCs", nil)
+		}
+	}
+
+	isMultiTenant := vpc1.TenantID != vpc2.TenantID
+
+	// Validate site access based on role
+	if isProviderAdmin {
+		// Provider Admin: org must have an infrastructure provider, and site must belong to that provider
```
thossain-nv: We also need to check that both Tenants have Tenant Account with the Provider.

api/pkg/api/handler/vpcpeering.go
```diff
@@ -0,0 +1,939 @@
...
+		logger.Error().Err(err).Msg("error updating VPC peering status to Ready after successful workflow")
+		return cutil.NewAPIErrorResponse(c, http.StatusInternalServerError, "Failed to update VPC peering status to Ready", nil)
+	}
+
+	// Update the API model to reflect the Ready status (workflow completed successfully)
+	apiVpcPeering := model.NewAPIVpcPeering(*vpcPeering)
+	apiVpcPeering.Status = cdbm.VpcPeeringStatusReady
+	logger.Info().Str("VPC Peering ID", vpcPeering.ID.String()).Msg("successfully created VPC peering and completed creation workflow")
+
+	// Commit the transaction
```
thossain-nv: The transaction needs to be committed immediately after Site workflow call. DB calls/updates after that should be on best effort basis (ok if fails).

thossain-nv (commented):
Thanks for the implementation @hanyux-nv Added some notes. Will review more tomorrow.

db/pkg/db/model/vpcpeering.go
thossain-nv: Each of the foreign keys should have the related object definition e.g.
Vpc1   *Vpc `bun:"rel:belongs-to,join:vpc_id=id"`

Example: https://github.com/NVIDIA/ncx-infra-controller-rest/blob/main/db/pkg/db/model/allocation.go#L123

This needed to make `includeRelations` param of `GetAll` functional.

db/pkg/db/model/vpcpeering.go
```diff
@@ -73,6 +73,9 @@ type VpcPeering struct {
 
 	IsMultiTenant bool `bun:"is_multi_tenant,notnull"`
 
+	InfrastructureProviderID *uuid.UUID `bun:"infrastructure_provider_id,type:uuid"`
```
thossain-nv: We need a migration for these additions.

db/pkg/db/model/vpcpeering.go
```diff
@@ -234,6 +249,27 @@ func (vpsd VpcPeeringSQLDAO) setQueryWithFilter(filter VpcPeeringFilterInput, qu
 		vpsd.tracerSpan.SetAttribute(vpDAOSpan, "site_ids", filter.SiteIDs)
 	}
 
+	if filter.InfrastructureProviderID != nil && filter.TenantID != nil {
```
thossain-nv: The `OR` logic only applies to the `VpcIDs` argument. For these two we want to handle it this way:

hwadekar-nv (approved):

PR #305, opened 2026-03-24
feat: Add full CRUD parity for TUI interactive mode

kfelternv:
## Summary

- Adds interactive create, update, and delete commands for the following resource types in the TUI: site, VPC, VPC virtualization, subnet, instance (create/delete), operating-system, SSH key group, SSH key, allocation, IP block, NSG, VPC prefix, and tenant account
- Scopes instance-create machine selection to the VPC's site so only relevant machines are presented, clearing any prior VPC scope during machine fetch to avoid stale filtering
- Expands scope-filtered cache invalidation to cover newly added resource types and fixes machine scoping to include VPC ID when set

## Test plan

- [x] `go build ./cli/...` passes
- [x] `go test ./cli/tui/... -count=1` -- all 25 tests pass
- [ ] Manual walkthrough of each new interactive command against a live or local Carbide instance
- [ ] Verify scope filtering applies correctly when site/VPC are set in the session

### DIFF
--- a/.github/workflows/build-push-service.yml
+++ b/.github/workflows/build-push-service.yml
@@ -125,6 +125,7 @@ jobs:
             TAGS="${TAGS},${{ inputs.target_registry }}/${{ inputs.service_name }}:${{ inputs.branch_sha_tag }}"
           fi
 
+          TAGS="${TAGS#,}"
           echo "tags=$TAGS" >> $GITHUB_OUTPUT
 
       - name: Build and push ${{ inputs.service_name }} (multi-arch)


kfelternv:
Completed a full re-review against current PR state (head `b66afae`, 4 files changed), and verified these findings are accurate against the live diff.

Findings (ordered by impact):

1) Instance create can be scoped by stale VPC scope
- In `cli/tui/session.go`, `fetchMachines` filters by `s.Scope.VpcID` when set.
- In `cmdInstanceCreate`, we resolve a selected VPC but only update site scope via `setSiteScopeFromID`; we do not set or clear `Scope.VpcID`.
- If session VPC scope was previously set, machine selection can be filtered by that stale scope instead of the VPC selected in the create flow.

2) Missing filtered cache invalidation for instance mutations
- `cmdInstanceCreate` and `cmdInstanceDelete` call `s.Cache.Invalidate("instance")` only.
- `instance` is in `InvalidateFiltered()` resource set in `cli/tui/cache.go`.
- This can leave stale scope-filtered cache entries after instance mutations.

3) PR description overstates implemented CRUD coverage
- Description says CRUD parity includes DPU extension service and implies all remaining resources.
- `cli/tui/commands.go` adds only `dpu-extension-service list/get` (no create/update/delete), and there is no `instance update` command.
- API support for DPU extension service CUD does exist (`openapi/spec.yaml` and `api/pkg/api/routes.go`).

Additional validated notes:
- `cmdTenantAccountUpdate` empty `{}` body is intentional per OpenAPI (`TenantAccountUpdateRequest` says no params needed).
- `cmdSSHKeyUpdate` and `cmdVPCPrefixUpdate` requiring name match their OpenAPI request schemas.
- `fetchMachinesWithSiteFallback` uses string-matching on `"400"`; functional but brittle.
- Prefix parsing via `fmt.Sscanf` is permissive for trailing characters.

Current PR state checks:
- CI is green across lint/tests/build checks.
- One inline review comment remains from CodeRabbit.

This comment intentionally excludes any claims that could not be confirmed from the live PR state.

PR #350, opened 2026-04-06
fix: Allow setting routing profile type when creating VPCs

bcavnvidia:
## Description
This change adds VPC routing profile selection to the REST API for VPC creation.

Callers can now include `routingProfileType` when creating a VPC.

`routingProfileType` is only accepted when `networkVirtualizationType` is FNN, and the API response only returns routingProfileType for that network virtualization type.

## Type of Change

## Services Affected

## Breaking Changes

## Testing

## Additional Notes
Satisfies https://github.com/NVIDIA/ncx-infra-controller-rest/issues/327


api/pkg/api/model/vpc.go
```diff
@@ -298,11 +341,13 @@ type APIVpc struct {
 	Created time.Time `json:"created"`
 	// Updated indicates the ISO datetime string for when the VPC was last updated
 	Updated time.Time `json:"updated"`
+	// RoutingProfileType is the requested routing profile type for the VPC.
+	RoutingProfileType *string `json:"routingProfileType"`
```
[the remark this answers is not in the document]
bcavnvidia: We prefer to explicitly return null.

thossain-nv: Could we move all 3 attributes at before `Status`? For most API models, Status, StatusHistory, Created, Updated are the last 4 attributes.

api/pkg/api/handler/vpc.go
```diff
@@ -249,6 +249,13 @@ func (cvh CreateVPCHandler) Handle(c echo.Context) error {
 		}
 	}
 
+	if apiRequest.RoutingProfileType != nil && *networkVirtualizationType != cdbm.VpcFNN {
+		logger.Warn().Str("routingProfileType", *apiRequest.RoutingProfileType).Msg("`routingProfileType` is only supported when `networkVirtualizationType` is FNN")
+		return cutil.NewAPIErrorResponse(c, http.StatusBadRequest, "`routingProfileType` is only supported when `networkVirtualizationType` is FNN", validation.Errors{
+			"routingProfileType": errors.New("`routingProfileType` is only supported when `networkVirtualizationType` is FNN"),
```
thossain-nv: When user doesn't explicitly specify `networkVirtualizationType`, we infer it from Site native networking flag. So we'll have to modulate this error based on whether `apiRequest.NetworkVirtualizationType` is set or not.

thossain-nv (commented):
Thanks for the implementation @bcavnvidia Left some notes.

api/pkg/api/model/vpc.go
```diff
@@ -91,9 +120,23 @@ func (ascr APIVpcCreateRequest) Validate() error {
...
+	if ascr.RoutingProfileType != nil {
+		if _, ok := VpcRoutingProfileTypeProtobufFromAPI[*ascr.RoutingProfileType]; !ok {
+			return validation.Errors{
+				"routingProfileType": fmt.Errorf("unsupported `routingProfileType`: %s", *ascr.RoutingProfileType),
+			}
+		}
+
+		if ascr.NetworkVirtualizationType != nil && *ascr.NetworkVirtualizationType != cdbm.VpcFNN {
+			return validation.Errors{
+				"routingProfileType": errors.New("`routingProfileType` is only supported when `networkVirtualizationType` is FNN"),
```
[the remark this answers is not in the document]
bcavnvidia: This one here would only be caught if they had specified the nw virt type.

This gets caught in two spots.  If nw virt were specified, and it's not FNN, then this triggers.  In the handler, after the nw virt has been fully resolved, then we test again and can be more specific.  This one here seems ok as it is, but I think the one in the handler could probably use the proposed wording you have for this one.

api/pkg/api/model/vpc.go
```diff
@@ -91,9 +120,23 @@ func (ascr APIVpcCreateRequest) Validate() error {
 		}
 	}
 
+	if ascr.RoutingProfileType != nil {
+		if _, ok := VpcRoutingProfileTypeProtobufFromAPI[*ascr.RoutingProfileType]; !ok {
+			return validation.Errors{
+				"routingProfileType": fmt.Errorf("unsupported `routingProfileType`: %s", *ascr.RoutingProfileType),
```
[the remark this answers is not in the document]
bcavnvidia: We might be very close to switching this over to being a purely config/db-defined string.  I.e., `external` might be a required minimum, but any number of profiles could be defined, and the proto would switch to being a string. I was trying to strike a middle-ground since we do need to expose this now.

thossain-nv (commented):
Thanks for the changes @bcavnvidia One more question I have, if routing profile type is not specified, should we set a default value for FNN VPCs?

api/pkg/api/model/vpc.go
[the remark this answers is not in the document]
bcavnvidia: Profiles are now string-backed and dependent on -core config.  No more predefined/enum profiles.

rla/internal/carbideapi/carbideproto/forge.proto
```diff
@@ -4427,14 +4432,16 @@ message Tenant {
...
+  optional string routing_profile_type = 5;
 }
 
 message CreateTenantRequest {
   // this organization ID must be globally unique
   string organization_id = 1;
   Metadata metadata = 2;
-  optional RoutingProfileType routing_profile_type = 3;
+  reserved 3;
+  optional string routing_profile_type = 4;
```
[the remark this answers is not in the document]
bcavnvidia: @coderabbitai This is known and expected.  This proto spec is owned by a different repo, and _this_ repo is the only official consumer of these fields _after_ this PR is merged.  We're reserving the fields as a matter of process, but there is no need to support a deprecated field in this specific case.

thossain-nv:
@bcavnvidia A few questions:
- How would the user determine what values they can specify for routing profiles? Seems like we took away fixed values.
-  When nothing is specified, would the routing profile type be `null`? Can routing profile type be updated after VPC creation?

bcavnvidia:
> @bcavnvidia A few questions:
> 
>     * How would the user determine what values they can specify for routing profiles? Seems like we took away fixed values.
> 
>     * When nothing is specified, would the routing profile type be `null`? Can routing profile type be updated after VPC creation?


- For VPCs, routing-profile is on creation only.
- It purely depends on what profiles are defined in the -core config.  I think we'd eventually either expose an API to manage profiles or store the details in the runtime_config details that come back from the version call to -core and sync that info to -rest.

thossain-nv:
Without a way for user to understand what profiles are available, adding this attribute seems incomplete. Can we add the currently supported routing profile type in API model as constant or in OpenAPI spec while we figure out how to propagate them?

bcavnvidia:
Seems I only answered part of your earlier question.

> When nothing is specified, would the routing profile type be `null`?

If the routing-type is omitted, the VPC simply inherits the profile of the tenant, which is what has been happening silently up until now.

_Tenant_ routing-profile can also be specified on tenant creation and update (as long as the tenant has no VPCs) in -core, but that can _also_ be omitted on creation and will get defaulted to a value specified in config, which defaults to `EXTERNAL` to match the behavior up to now,  I.e., if no default tenant profile is specified in config, no profile is sent on creation or updated after creation, and no profile is sent in on VPC creation, things will function exactly as they do now:  tenant is created as `EXTERNAL` and their VPCs are created with that profile, and the definition of that `EXTERNAL` profile is then required in the -core config.  We have [documentation](https://github.com/NVIDIA/ncx-infra-controller-core/blob/main/book/src/manuals/vpc/vpc_routing_profiles.md) to direct users when the site is lacking configuration. (Which is getting an [update](https://github.com/NVIDIA/ncx-infra-controller-core/pull/1079) based on the new behavior.)

> Without a way for user to understand what profiles are available, adding this attribute seems incomplete. Can we add the currently supported routing profile type in API model as constant or in OpenAPI spec while we figure out how to propagate them?

The prior, fixed set of names didn't guarantee that the profiles were truly available, since it still depended on what was actually defined in the -core config.   Even then, it depends on the profile being requested vs the profile of the tenant, since a tenant can't request profiles with "more access" then their own profile when creating a VPC. It's similar to explicit VNI requests (and provider-level IP blocks), where available VNIs would only be known if you have knowledge of how the site were configured.

This option isn't really meant for typical tenants.  I.e., it's not really meant as a true user-facing option; rather, it's a service-facing option. It's being plumbed through -rest because -rest is now the official service-to-service API as well, and it'd probably be ideal if we could restrict the option to tenants based on privileges.

EDIT: Seems like maybe what we want is to gate this to "privileged tenants" only?

EDIT 2:
Decided we _will_ restrict it for now, and we'll also restrict it to some "well-known" names but leave the API contract and handling able to simply open up to free-form later.  I'll update the NICo -core docs to steer people toward a minimum config that includes the well-known names.

openapi/spec.yaml
```diff
@@ -13100,6 +13106,11 @@ components:
             - ETHERNET_VIRTUALIZER
             - FNN
           description: 'Network virtualization type of the VPC. If no value is specified, then defaults to `FNN` if Site has native networking enabled, or ETHERNET_VIRTUALIZER if native networking is disabled'
+        routingProfileType:
+          type:
+            - string
+            - 'null'
+          description: Routing profile type to request for the VPC. Only supported when `networkVirtualizationType` is `FNN`
```
thossain-nv: Suggested description:
```
Specify routing profile for the VPC. Only supported when `networkVirtualizationType` is set to `FNN`, or when `networkVirtualizationType` is omitted and Site has Native Networking enabled. Requires Tenant to have elevated privilege. Current accepted values are `privileged-internal`, `internal`, and `external`
```

openapi/spec.yaml
```diff
@@ -12985,6 +12985,11 @@ components:
             - ETHERNET_VIRTUALIZER
             - FNN
           description: Network virtualization type of the VPC
+        routingProfileType:
+          type:
+            - string
+            - 'null'
+          description: Requested routing profile type for the VPC. Only supported when `networkVirtualizationType` is `FNN`
```
thossain-nv: Suggested description:
```
Routing profile type for the VPC. Populated when Site has Native Networking enabled and network virtualization type is `FNN` 
```

api/pkg/api/handler/vpc.go
```diff
@@ -249,6 +249,25 @@ func (cvh CreateVPCHandler) Handle(c echo.Context) error {
...
+
+		if *networkVirtualizationType != cdbm.VpcFNN {
+			logger.Warn().Str("routingProfile", *apiRequest.RoutingProfile).Msg("`routingProfile` can only be specified if network virtualization type is set to `FNN`, or Site has native networking enabled and no network virtualization type is specified")
+			return cutil.NewAPIErrorResponse(c, http.StatusBadRequest, "`routingProfile` can only be specified if network virtualization type is set to `FNN`, or Site has native networking enabled and no network virtualization type is specified", nil)
+		}
+	}
+
+	var routingProfile *string
+	if apiRequest.RoutingProfile != nil {
+		routingProfile = cdb.GetStrPtr(model.NormalizeAPIVpcRoutingProfileForSite(*apiRequest.RoutingProfile))
```
thossain-nv, line 267: We could put this statement within the above conditional after line 263?

api/pkg/api/model/vpc.go
```diff
@@ -246,6 +317,12 @@ func NewAPIVpc(dbVpc cdbm.Vpc, dbsds []cdbm.StatusDetail) APIVpc {
 
 	if dbVpc.NetworkVirtualizationType != nil {
 		apivpc.NetworkVirtualizationType = dbVpc.NetworkVirtualizationType
+		if *dbVpc.NetworkVirtualizationType == cdbm.VpcFNN {
```
thossain-nv: Do we need this additional conditional? Setting this is only for FNN gated by the API. If it's populated, we should return it, no?

thossain-nv (approved):
Thanks @bcavnvidia, a few final suggestions.

api/pkg/api/handler/vpc.go
bcavnvidia: Oh, yeah, funny, I moved one but missed this one.  Thanks!

PR #371, opened 2026-04-09
fix: Remove deprecated instance/allocation relationships

bcavnvidia:
## Description
This change removes the legacy per-instance allocation linkage from the REST/API and database layers and finishes the transition to enforcement based on aggregate allocation.

- Instance creation and batch creation now compare the total reserved capacity available for a tenant’s instance type at a site
- allocation-constraint updates and  allocation deletion checks validate against total instance usage instead of usage attached to a single allocation record.

## Type of Change

## Services Affected

## Breaking Changes

## Testing


api/pkg/api/handler/instance.go
```diff
@@ -1256,12 +1231,10 @@ func (cih CreateInstanceHandler) Handle(c echo.Context) error {
...
 
-	// NOTE: Set InstanceTypeID only if it is provided in the request. For Instances created with an Instance Type ID, we expect Allocation information
-	// to be present. Since Machine ID based Instance creation does not require Allocation information, setting InstanceTypeID will create data integrity issues.
+	// NOTE: Set InstanceTypeID only if it is provided in the request.
+	// Machine ID based Instance creation does not require an Instance Type.
 	if apiRequest.InstanceTypeID != nil {
 		instanceCreateInput.InstanceTypeID = instanceTypeID
-		instanceCreateInput.AllocationID = &selectedAllocationConstraint.AllocationID
-		instanceCreateInput.AllocationConstraintID = &selectedAllocationConstraint.ID
 	}
```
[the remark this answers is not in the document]
bcavnvidia: I believe this is actually correct.  If a machine ID is provided instead of an instance type ID, then this is a "targeted" instance creation and is not supposed to count against allocation limits.

@thossain-nv

api/pkg/api/handler/allocation.go
```diff
@@ -1277,9 +1284,38 @@ func (dah DeleteAllocationHandler) Handle(c echo.Context) error {
...
+			if serr != nil {
+				logger.Error().Err(serr).Msg("error getting Allocation Constraint value for Allocation being deleted")
+				return cutil.NewAPIErrorResponse(c, http.StatusInternalServerError, "Error getting Allocation capacity for Allocation being deleted", nil)
+			}
+
+			// If deleting this allocation would reduce the remaining aggregate capacity
+			// below the current instance count, then the delete must be blocked.
+			if iCount > totalConstraintValue-deletedConstraintValue {
+				logger.Warn().Msg("unable to delete Allocation, existing instances would exceed the remaining pool capacity")
+				return cutil.NewAPIErrorResponse(c, http.StatusBadRequest, fmt.Sprintf("%v existing Instances in the tenant/site/instance type pool would exceed the remaining capacity after deleting this Allocation", iCount), nil)
```
bcavnvidia, line 1318: Probably should be something like
`"%d existing Instances would exceed the total allocation capacity after deleting this Allocation"`

api/pkg/api/handler/instance.go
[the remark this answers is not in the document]
bcavnvidia: @coderabbitai I've confirmed that my explanation above is correct.

api/pkg/api/handler/instance.go
```diff
@@ -944,51 +940,30 @@ func (cih CreateInstanceHandler) Handle(c echo.Context) error {
 		if instanceType.SiteID != nil {
 			siteIDs = []uuid.UUID{*instanceType.SiteID}
 		}
-		instances, insTotal, err := instanceDAO.GetAll(ctx, tx, cdbm.InstanceFilterInput{TenantIDs: []uuid.UUID{tenant.ID}, SiteIDs: siteIDs, InstanceTypeIDs: []uuid.UUID{instanceType.ID}}, cdbp.PageInput{Limit: cdb.GetIntPtr(cdbp.TotalLimit)}, nil)
+		_, insTotal, err := instanceDAO.GetAll(ctx, tx, cdbm.InstanceFilterInput{
+			TenantIDs:       []uuid.UUID{tenant.ID},
+			SiteIDs:         siteIDs,
+			InstanceTypeIDs: []uuid.UUID{instanceType.ID},
+		}, cdbp.PageInput{}, nil)
```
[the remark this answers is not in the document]
bcavnvidia: @coderabbitai "active" in this context really means "exists."  If an instance is not deleted, in which case it would not be returned in the query, then it must be counted against allocation numbers.

api/pkg/api/handler/util/common/common.go
```diff
@@ -303,6 +303,17 @@ func GetAllocationIDsForTenantAtSite(ctx context.Context, tx *cdb.Tx, dbSession
 	return aIDs, nil
 }
 
+// AcquireInstanceTypeQuotaLock acquires the shared quota lock for a tenant/site/instance-type pool.
+// It returns an error if the advisory lock cannot be acquired.
+func AcquireInstanceTypeQuotaLock(ctx context.Context, tx *cdb.Tx, tenantID uuid.UUID, siteID uuid.UUID, instanceTypeID uuid.UUID) error {
```
bcavnvidia: Site ID is technically a property of instance type, so we should be able to key purely on tenant and instance type, but there are a few places that might make people assume otherwise:

  - api/pkg/api/handler/instance.go
      - the active-instance count uses InstanceFilterInput{TenantIDs, SiteIDs, InstanceTypeIDs}
      - SiteIDs is populated from instanceType.SiteID
  - api/pkg/api/handler/instancebatch.go
      - same pattern: TenantIDs, SiteIDs, and InstanceTypeIDs
      - again, siteIDs is derived from instancetype.SiteID
  - api/pkg/api/handler/allocationconstraint.go
      - the aggregate validation counts instances with TenantIDs, SiteIDs, and InstanceTypeIDs
  - api/pkg/api/handler/allocation.go
      - the delete guard also counts instances with TenantIDs, SiteIDs, and InstanceTypeIDs

I think we can clean all of that up in a follow-up PR.

db/pkg/migrations/20260409143000_remove_instance_allocation_relationship.go
```diff
@@ -0,0 +1,66 @@
+/*
```
thossain-nv, line 1: I'd defer the actual column removal in a subsequent PR. We can first remove the DB attributes and their usage everywhere. Then the migration can be run while previous version of API pods are still serving Instance creation.

api/pkg/api/handler/allocation.go
thossain-nv: Better as:
```go
fmt.Sprintf("Deleting this Allocation as specified would result in %d total Machines for Instance Type: %s allocated to Tenant, less than Tenant's active Instance count: %d for the Instance Type", totalConstraintValue-deletedConstraintValue, instanceType.Name, iCount)
```

api/pkg/api/handler/allocationconstraint.go
```diff
@@ -206,26 +206,65 @@ func (uach UpdateAllocationConstraintHandler) Handle(c echo.Context) error {
 				return cutil.NewAPIErrorResponse(c, http.StatusBadRequest, "Error retrieving Instance Type in Allocation Constraint in request", nil)
 			}
 
-			// acquire an advisory lock on the InstanceType
-			// this lock is released when the transaction commits or rollsback
-			serr = tx.TryAcquireAdvisoryLock(ctx, cdb.GetAdvisoryLockIDFromString(dbit.ID.String()), nil)
+			// Acquire the shared quota lock for this tenant/site/instance-type pool.
+			// This lock is released when the transaction commits or rolls back.
+			serr = common.AcquireInstanceTypeQuotaLock(ctx, tx, a.TenantID, dbit.ID)
```
thossain-nv, line 211: The reason this was an Instance Type wide lock was to prevent simultaneously over allocating the same Instance Type to 2 or more Tenants. The Tenant/Instance Type ID lock circumvents that, no?

api/pkg/api/handler/allocationconstraint.go
```diff
@@ -206,26 +206,65 @@ func (uach UpdateAllocationConstraintHandler) Handle(c echo.Context) error {
...
 				logger.Error().Err(serr).Msg("error getting Instances from db for AllocationConstraint")
 				return cutil.NewAPIErrorResponse(c, http.StatusInternalServerError, "Failed to retrieve Instances for AllocationConstraint", nil)
 			}
-			if iCount > apiRequest.ConstraintValue {
+
+			// If the request is reducing the constraint value, make sure the total
+			// pool does not fall below what has already been allocated.
+			if iCount > sumConstraints {
 				logger.Warn().Msg("current number of Instances exceed Constraint value, failed to update Allocation Constraint")
 				return cutil.NewAPIErrorResponse(c, http.StatusBadRequest, "Cannot update AllocationConstraint, more Instances exist than requested constraint for this Allocation", nil)
```
thossain-nv, line 263: Similar to informative message in a previous comment:

```go
fmt.Sprintf("Updating this Allocation as specified would result in %d total Machines for Instance Type: %s allocated to Tenant, less than Tenant's active Instance count: %d for the Instance Type", sumConstraints, instanceType.Name, iCount)
```

thossain-nv (commented):
Thanks for this @bcavnvidia A few suggestions.

api/pkg/api/handler/allocationconstraint.go
bcavnvidia: I think this was a coderabbit+codex fix that got slipped in, but it actually does seem to be a partially correct fix because Instance creation coordinates on tenant+type.

AllocationConstraint delete, AllocationConstraint update for _decrease_, and Instance creation should coordinate around the same thing; otherwise, a tenant could create an instance at the same time their allocation was being decreased.

What should happen here is coordination around tenant+type if the update is a decrease, and coordination should be around only instance-type if the update is an increase.

The following should all coordinate around  only instance-type:
- Machine removal from instance-type
- Allocation created
- Allocation increased

The following should all coordinate around tenant+type:
- Allocation delete
- Allocation decrease
- Instance creation

bcavnvidia: Have another look at the changes in this file.  It starts with the granular lock and "upgrade" if it needs to.

thossain-nv: Thanks Brian for the detailed analysis. Going through the code again.

thossain-nv (approved):
Thanks for this @bcavnvidia, I think we have covered all cases.

PR #379, opened 2026-04-13
fix: Skip config filter in DB if no config query params are set when retrieving all Sites

kfelternv:
## Summary

- The GetAllSite handler unconditionally set `filter.Config` to an empty `SiteConfigFilterInput`, causing the DAO to always apply `st.config @> '{}'::jsonb`. Sites with a NULL config column evaluate NULL for this predicate and are silently excluded from results, even when no config filtering was requested.
- This caused `carbidecli site list` to return 0 items while `site create` could still detect the existing site via a 409 conflict (reported in reno-qa4).
- Only set `filter.Config` when at least one config query parameter (`isNativeNetworkingEnabled`, `isNetworkSecurityGroupEnabled`, `isNVLinkPartitionEnabled`, `isRackLevelAdministrationEnabled`) is actually provided.

## Test plan

- Added `TestGetAllSiteHandler_NullConfig` that reproduces the exact NVBug scenario: creates a site named "reno-qa4", sets its config column to NULL via direct SQL, then calls both `GetAllSiteHandler` (returns 0 items) and `CreateSiteHandler` (returns 409). Before the fix the test fails with the same contradiction as the bug report; after the fix both handlers agree the site exists.
- All existing `TestGetAllSiteHandler_Handle` subtests continue to pass.
- Still need to verify the fix against the live environment where the bug was reported, pending access to the correct NGC org and cloud API credentials.


api/pkg/api/handler/site.go
```diff
@@ -779,55 +779,56 @@ func (gash GetAllSiteHandler) Handle(c echo.Context) error {
...
 	// Check `isNativeNetworkingEnabled` in query
-	var isNativeNetworkingEnabled *bool
 	qinne := c.QueryParam("isNativeNetworkingEnabled")
 	if qinne != "" {
 		isnnEnabled, err := strconv.ParseBool(qinne)
 		if err != nil {
 			return cutil.NewAPIErrorResponse(c, http.StatusBadRequest, "Invalid value specified for `isNativeNetworkingEnabled` query param", nil)
 		}
-		isNativeNetworkingEnabled = &isnnEnabled
+		configFilter.NativeNetworking = &isnnEnabled
```
thossain-nv, line 792: Perhaps we rename `isnnEnabled` to `isEnabled` to be consistent with the other conditionals.

thossain-nv (approved):
Thanks for the fix @kfelternv

PR #383, opened 2026-04-14
fix: Maintain association record when Instance Type is updated in Machine inventory

thossain-nv:
## Description

Currently we're only updating Instance Type attribute for Machine and not the Machine/Instance Type association record. Until Machine/Instance Type association is deprecated, we need to keep it consistent.

## Type of Change

## Services Affected

## Related Issues (Optional)

None

## Breaking Changes

## Testing

## Additional Notes

None

### DIFF
--- a/workflow/pkg/activity/machine/machine.go
+++ b/workflow/pkg/activity/machine/machine.go
@@ -217,6 +217,7 @@ func (mm *ManageMachine) UpdateMachinesInDB(ctx context.Context, siteIDStr strin
 	}
 
 	miDAO := cdbm.NewMachineInterfaceDAO(mm.dbSession)
+	mitDAO := cdbm.NewMachineInstanceTypeDAO(mm.dbSession)
 	sdDAO := cdbm.NewStatusDetailDAO(mm.dbSession)
 
 	// Iterate through Machines in the inventory and update/create them in DB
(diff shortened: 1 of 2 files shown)


workflow/pkg/activity/machine/machine.go
```diff
@@ -460,38 +461,20 @@ func (mm *ManageMachine) UpdateMachinesInDB(ctx context.Context, siteIDStr strin
...
-					InstanceTypeID:       clearInstanceTypeID,
-				}
-				_, serr = mDAO.Clear(ctx, txn, clearInput)
-				if serr != nil {
-					slogger.Error().Err(serr).Msg("failed to clear maintenance or network health message from Machine in DB")
-				}
-			}
+			// Check if there were any changes to the Instance type ID
+			clearInstanceTypeID := controllerInstanceTypeID == nil && existingCloudMachine.InstanceTypeID != nil
+			updateInstanceTypeID := controllerInstanceTypeID != nil && (existingCloudMachine.InstanceTypeID == nil || *controllerInstanceTypeID != *existingCloudMachine.InstanceTypeID)
```
bcavnvidia: You could use PtrsEqual here.

bcavnvidia (approved):

workflow/pkg/activity/machine/machine.go
thossain-nv: Good point, updating!

PR #385, opened 2026-04-14
fix: Have PSM read the firmware files at startup time rather than using an embedded filesystem

spydaNVIDIA:
## Description

fix: Have PSM read the firmware files at startup time rather than using an embedded filesystem

## Type of Change

## Services Affected

- [x] **Powershelf Manager** - Powershelf Manager updated

## Breaking Changes

## Testing


powershelf-manager/pkg/redfish/client.go
```diff
@@ -39,6 +40,20 @@ import (
 	"github.com/stmcginnis/gofish/redfish"
 )
 
+// checkResponse returns an error if the HTTP status code indicates failure (>= 300),
+// including the response body in the error message for diagnostics.
+func checkResponse(resp *http.Response) error {
+	if resp.StatusCode < http.StatusMultipleChoices {
+		return nil
+	}
+	defer resp.Body.Close()
```
kunzhao-nv: I think we should move this line forward, otherwise it leaks HTTP connections on success.

powershelf-manager/pkg/firmwaremanager/firmware_updater.go
```diff
@@ -147,21 +151,21 @@ func (fp *FirmwareUpdater) update(client *redfish.RedfishClient, targetVersion f
 
 	log.Printf("FW Updater does not support updating powershelf that has a PMC fw version of r.%v.%v.%v\n", currentVersion.major, currentVersion.minor, currentVersion.patch)
 
-	return nil, nil
+	return nil
```
kunzhao-nv: When `canUpdate` is false or `getFwUpgrade` returns nil, the function logs and returns nil. In the caller `handleOnePmcUpdate`, this means the state transitions to FirmwareStateVerifying even though no upload was actually performed. The firmware update will then sit in "verifying" for an hour until the timeout handler marks it as failed.

I think we should return error here, like
```
return fmt.Errorf("no supported firmware upgrade path from %s for vendor %s",
    currentVersion.String(), fp.vendor.String())

File: powershelf-manager/pkg/firmwaremanager/firmware_repo.go
Comment (kunzhao-nv) [inline powershelf-manager/pkg/firmwaremanager/firmware_repo.go]: When updating max we should compare with max?
=>
```
from.cmp(maxStartingFwVersion) > 0

powershelf-manager/cmd/firmware_manager.go
```diff
@@ -83,32 +83,26 @@ func init() {
 	fwCmd.Flags().StringVarP(&pmcUsername, "user", "u", "root", "Username")
 	fwCmd.Flags().StringVarP(&pmcPassword, "pass", "p", "0penBmc", "Password")
 	fwCmd.Flags().StringVar(&versionTo, "version", "0penBmc", "Target Version to upgrade to")
+	fwCmd.Flags().StringVarP(&pmcMAC, "mac", "m", "", "PMC MAC address")
+	fwCmd.Flags().StringVar(&firmwareDir, "firmware-dir", getEnvOrDefault("FW_DIR", "/var/lib/psm/firmware"), "Firmware files directory (env: FW_DIR)")
```
kunzhao-nv: nit:
here is `firmware-dir` but serve.go is `fw_dir`. Better make them consistency.

powershelf-manager/cmd/firmware_manager.go
```diff
@@ -83,32 +83,26 @@ func init() {
 	fwCmd.Flags().StringVarP(&pmcUsername, "user", "u", "root", "Username")
 	fwCmd.Flags().StringVarP(&pmcPassword, "pass", "p", "0penBmc", "Password")
 	fwCmd.Flags().StringVar(&versionTo, "version", "0penBmc", "Target Version to upgrade to")
```
kunzhao-nv: Why version default is `0penBmc`?

powershelf-manager/cmd/redfish.go
```diff
@@ -230,7 +232,10 @@ func doRedfish() {
 		}
 		util.PrintPrettyResponse(resp)
 	case UploadFirmwareFile:
```
kunzhao-nv, line 234: It still calls client.UploadFirmwareByPath() directly, skipping the new apply-time setup and checkResponse validation that UpdateFirmware now provides. This means a 4xx/5xx from the device just gets logs instead of failing the command. We we route through UpdateFirmware here too (open the file, pass an io.Reader).

powershelf-manager/cmd/firmware_manager.go
```diff
@@ -83,32 +83,26 @@ func init() {
...
 
-	pmc := &pmc.PMC{
-		IP:     ip,
-		Vendor: vendor,
-		Credential: &credential.Credential{
-			User:     pmcUsername,
-			Password: secretstring.New(pmcPassword),
-		},
+	cred := credential.New(pmcUsername, pmcPassword)
+	pmc, err := pmc.New(pmcMAC, pmcIP, vendor.Code, &cred)
```
kunzhao-nv: nit:
`pmc, xxx = pmc.xxx`:
I think we can assign the first "pmc" here a better name like `pmcInstance`

powershelf-manager/cmd/firmware_manager.go
```diff
@@ -40,6 +39,7 @@ var vendorStr string
 var fw_action string
```
kunzhao-nv: `fw_action`, `dryRun`
Do we want snake case or camel case here?

powershelf-manager/cmd/serve.go
```diff
@@ -137,6 +142,7 @@ func doServe() {
 				Credential:        credential.New(dbUser, dbPassword),
 				CACertificatePath: dbCertPath,
 			},
+			FirmwareDir: firmwareDir,
```
kunzhao-nv, line 145: We should add `firmwareDir` in log for debugging.

powershelf-manager/pkg/firmwaremanager/embedded_firmware.go
```diff
@@ -17,39 +17,35 @@
...
 	"strings"
 
+	"github.com/NVIDIA/ncx-infra-controller-rest/powershelf-manager/pkg/common/vendor"
 	log "github.com/sirupsen/logrus"
 )
 
-//go:embed firmware/*
-var firmware embed.FS
-
 const firmware_path = "firmware"
```
kunzhao-nv: With os.DirFS(firmwareDir) and the constant firmware_path = "firmware", the expected path is <FW_DIR>/firmware/<vendor>/pmc/<files>. If an operator sets FW_DIR=/var/lib/psm/firmware, the actual files live at /var/lib/psm/firmware/firmware/liteon/pmc/. This double-nesting may be confusing.

Consider either:
- Making firmwareDir the direct parent of vendor directories, or
- Documenting/Commenting the expected layout.

kunzhao-nv (approved):

powershelf-manager/cmd/firmware_manager.go
spydaNVIDIA: renamed to fw_dir here

powershelf-manager/cmd/redfish.go
spydaNVIDIA: I think this is OK. This command is supposed to be mainly used for debugging and printing the response would give us all we need. In the future if we really want `psm redfish UploadFirmwareFile` to handle errors appropriately, we can.

powershelf-manager/cmd/firmware_manager.go
spydaNVIDIA: fixed to camel case in accordance with golang convention

powershelf-manager/pkg/firmwaremanager/embedded_firmware.go
spydaNVIDIA: fixed

PR #389, opened 2026-04-15
chore: Fix idempotency issue for warning comments in Core proto format script

pbreton:
## Description
Make 'make core-proto-fmt' idempotent and avoid duplicating lines and blocks.

## Type of Change
- [X] **Internal** - Fix/change internal tooling

## Testing
- [X] Manual testing performed


thossain-nv (approved):
Thanks @pbreton

PR #391, opened 2026-04-15
ci: Add Github workflow to ensure Core protobuf is up to date with proto files

pbreton:
## Description
chore: Add PR test to ensure generated code match current proto files

## Type of Change
- [X] **CI** - Changes in GitHub workflows. Requires additional scrutiny (ci:)

## Testing
- [X] Manual testing performed

### DIFF
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -160,6 +160,49 @@ jobs:
           fi
 
   # ============================================================================
+  # Check Protobuf Generated Code - Ensure .pb.go files match .proto sources
+  # ============================================================================
+  check-protobuf-generated:
+    name: Check Protobuf Generated Code
+    runs-on: ubuntu-latest
+    continue-on-error: false
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/lint-and-test.yml
```diff
@@ -159,6 +159,56 @@ jobs:
...
+          go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@latest
+
+      - name: Regenerate protobuf code
+        run: |
+          cd workflow-schema && buf generate
+          cd rla && buf generate
+
+      - name: Check for uncommitted changes
+        run: |
+          STATUS=$(git status --porcelain)
```
thossain-nv: Could we ignore generator version only changes or pin recommended versions?

pbreton: Probably ignoring is better for now otherwise we start enforcing version each developer uses...

hwadekar-nv (approved):

PR #401, opened 2026-04-17
fix: Prevent duplicate --data flag panic on carbidecli create commands

kfelternv:
## Summary

`carbidecli dpu-extension-service create` (and any future command whose request body declares a property named `data`, `data-file`, `output`, or `all`) panics before it ever runs:

```
create flag redefined: data
panic: create flag redefined: data
    ...
    github.com/urfave/cli/v2.(*StringFlag).Apply
    github.com/urfave/cli/v2.flagSet
    github.com/urfave/cli/v2.(*Command).newFlagSet
```

Root cause is in `buildActionCommand` in `cli/pkg/commands.go`: when the operation has a request body, the function registers hardcoded `--data` and `--data-file` flags, then iterates `schema.Properties` and registers one flag per property via `toKebab(name)`. `DpuExtensionServiceCreateRequest` has a property named `data`, so Go's stdlib `flag.StringVar` is called twice with the same name and panics at command setup time.

## Fix

Define a set of reserved flag names owned by the CLI wrapper (`output`, `all`, `data`, `data-file`). When a body property's kebab-cased name collides, register it under a `body-` prefix (e.g. `--body-data`) instead of creating a duplicate flag. The prefixed flag maps back to the original JSON property name in the request body, so no functionality is lost.

## Tests

Three unit tests in `cli/pkg/commands_test.go` (migrated to testify assert/require):

- `TestBuildCommands_NoDuplicateFlags` parses the embedded OpenAPI spec, walks every generated command and subcommand, and fails if any command declares the same flag name twice.
- `TestBuildActionCommand_ReservedBodyPropertyPrefixed` builds a synthetic spec with body properties `name`, `data`, `output`, `all` and asserts colliding properties get a `body-` prefix while non-colliding properties stay unprefixed.
- `TestNewApp_DpuExtensionServiceCreate_DoesNotPanic` runs the real spec through `app.Run` with `dpu-extension-service create --help`.

## Validation

- `go test ./cli/pkg/...` passes all tests.
- `go vet ./cli/...` and `go fmt ./cli/...` clean.
- Built binary against a local echo server, verified all request paths:
  - `--body-data` with individual flags sends `{"data":"...","name":"...","serviceType":"...","siteId":"..."}` -- prefixed flag correctly maps to `"data"` in JSON.
  - `--data '{"full":"json"}'` passthrough works unchanged.
  - `--data-file` passthrough works unchanged.
  - `--data` wrapper overrides individual flags including `--body-data` (early return in `buildRequestBody`).
  - Non-colliding commands (e.g. `site create`) work normally with no regressions.


cli/pkg/commands.go
```diff
@@ -324,11 +337,21 @@ func buildActionCommand(spec *Spec, ro resolvedOp, subResource string) *cli.Comm
 				if resolved == nil || resolved.Type == "object" || resolved.Type == "array" {
 					continue
 				}
+				flagName := toKebab(name)
+				// Skip body properties whose kebab-cased name collides with
```
thossain-nv: Do we have a body property in a request input called `data`?

thossain-nv (approved):
Looks good, one question.

cli/pkg/commands.go
kfelternv: _(AI Gen)_ Yes, DpuExtensionServiceCreateRequest has a `data` property (type: string, "Deployment spec for the DPU Extension Service"). It's defined in the OpenAPI spec under `components/schemas/DpuExtensionServiceCreateRequest` and maps to `Data string `json:"data"`` in `sdk/standard/model_dpu_extension_service_create_request.go`. That's the field colliding with the hardcoded `--data` flag in `buildActionCommand`, which triggers the panic on `carbidecli dpu-extension-service create`.

PR #405, opened 2026-04-20
fix: Move machine ID lock acquisition before record pull

bcavnvidia:
## Description
We're grabbing a lock on Machine.ID when allocating an instance, but it's happening after the machine record has already been pulled from the DB, and the locking helper being used has retries.  This means a retry could acquire the lock, but the checks that follow could be on stale data.

We either need to use the helper that fails fast, or we need the lock acquisition moved so that the lock happens before the record is pulled.  I did the latter.

Instance batch creation _seems_ to also expect the Try locking helper to fail fast, but the machine records are being re-pulled after lock acquisition there, so it's fine as it is.

## Type of Change

## Services Affected

## Breaking Changes

## Testing

### DIFF
--- a/api/pkg/api/handler/instance.go
+++ b/api/pkg/api/handler/instance.go
@@ -816,6 +816,13 @@ func (cih CreateInstanceHandler) Handle(c echo.Context) error {
 
 		mDAO := cdbm.NewMachineDAO(cih.dbSession)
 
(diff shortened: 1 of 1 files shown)


thossain-nv (approved):
Thanks for identifying and fixing this @bcavnvidia

PR #406, opened 2026-04-20
feat: Hint at label filter syntax after TUI list output

kfelternv:
# Summary

This change will make it easier to see how to compose commands in the TUI by displaying hints on how to run commands that filter resources by labels.

# AI GENERATED 

After running a list command in the TUI, users see a `LABELS` column populated with values like `RackIdentifier=H19`, `ServerName=pdx01-m01-h19-cpu-2`, but nothing on screen tells them they can filter or sort against those keys. The functionality already exists (`--label key=value`, `--sort-label key`, and persistent `scope label key=value`) but is only visible in `help`, which most people only run once.

Adding a one-line hint right where the labels appear closes the discovery gap without growing the help text or adding new commands.

## What

- New helpers in `cli/tui/commands.go`:
  - `uniqueLabelKeys` collects the sorted set of label keys present on a list of items
  - `printLabelHint` writes a one-line dim hint to stderr showing the syntax for `--label`, `--sort-label`, and `scope label`, using a real key from the result and listing up to six available label keys
- The hint is suppressed when any label filter is already active (scope or per-command), so it only surfaces when it is actionable
- Wired into the seven list commands that render a `LABELS` column: `vpc`, `instance-type`, `instance`, `machine`, `network-security-group`, `expected-machine`, `infiniband-partition`
- `help` now documents `<list> --label key=value` and `<list> --sort-label key`, which were previously only discoverable by reading the source

Hint goes to stderr to keep the table on stdout pipeable.

## Example

Before:

```
carbide:org/pdx-qa2-new> machine list
31 items
NAME                       STATUS         SITE          VPC   LABELS                                                    ID
pdx01-m01-h19-cpu-2        Error          pdx-qa2-new   -     RackIdentifier=H19, ServerName=pdx01-m01-h19-cpu-2        fm100ht...
...
```

After:

```
carbide:org/pdx-qa2-new> machine list
31 items
NAME                       STATUS         SITE          VPC   LABELS                                                    ID
pdx01-m01-h19-cpu-2        Error          pdx-qa2-new   -     RackIdentifier=H19, ServerName=pdx01-m01-h19-cpu-2        fm100ht...
...
Hint: --label RackIdentifier=<value> filter | --sort-label RackIdentifier sort | scope label RackIdentifier=<value> persist
Label keys: RackIdentifier, ServerName, test_key
```

Once the user runs `--label RackIdentifier=H19` or `scope label RackIdentifier=H19`, the hint disappears for subsequent list calls until the filter is cleared.

## Tests

- `TestUniqueLabelKeys` covers empty input, no-label items, dedupe, and sort
- `TestPrintLabelHint` covers active-filter suppression, no-labels suppression, key surfacing, key-list truncation past six keys, and the empty-vs-nil filter map distinction
- Full `cli/tui` and `cli/pkg` test suites pass locally

## Test plan

- [ ] `go test ./cli/...`
- [x] Manual: run `carbidecli i` against a stage env, run `machine list`, confirm the hint appears
- [x] Manual: run `scope label <key>=<value>` then `machine list`, confirm the hint is suppressed
- [x] Manual: run `vpc list`, `instance-type list`, `instance list`, `network-security-group list`, `expected-machine list`, `infiniband-partition list`, confirm hint behavior is consistent across all seven


cli/tui/commands.go
```diff
@@ -2426,6 +2435,58 @@ func formatLabels(labels map[string]string, maxWidth int) string {
...
+	keys := uniqueLabelKeys(items)
+	if len(keys) == 0 {
+		return
+	}
+	sample := keys[0]
+	fmt.Fprintln(w, Dim(fmt.Sprintf(
+		"Hint: --label %s=<value> filter | --sort-label %s sort | scope label %s=<value> persist",
+		sample, sample, sample,
+	)))
+	const maxKeys = 6
```
thossain-nv: Values like this `maxKeys: 6` might be better off being a well defined constant at the top of the file so we have a set of values that are tunable without diving too deep in the code.

thossain-nv (commented):
How does filtering/sorting by label work with pagination? The API currently supports free text searching by label key value using query param `query`

cli/tui/commands.go
kfelternv: Sounds good,

The pagination happens automatically, the TUI retrieves all pages. Its a naive approach and might need to have a hard upper limit when trying to list tens of thousands of resources.

thossain-nv: Filtering on the full set of pages is still helpful since we're fetching them. Agreed that it would be good to have a limit placed on the number of pages fetched. We can always suggest various filters to trim down the list to help user work with a smaller set of objects to begin with.

thossain-nv (approved):
Thanks @kfelternv

PR #408, opened 2026-04-21
fix: Resolve required query param for tenant-account list in TUI

kfelternv:
## Summary

- The `tenant-account list` command in TUI/interactive mode was failing with `API error 400: Either infrastructureProviderId or tenantId query param must be specified` because the fetcher passed no query params
- Auto-resolve the infrastructure provider ID via `infrastructure-provider/current` (or fall back to tenant ID via `tenant/current`) and pass it as a query param in the fetcher
- Include the resolved ID in the logged `carbidecli` command for reproducibility

This fix applies to all TUI commands that use the `tenant-account` fetcher (list, get, update, delete).

## Test plan

- Run `carbidecli tui` and execute `tenant-account list` -- should return results instead of a 400 error
- Verify the INFO line includes `--infrastructure-provider-id <id>` (or `--tenant-id <id>` for tenant-only orgs)
- Run `tenant-account get` from TUI to confirm the resolver-level fix works for downstream commands

### DIFF
--- a/cli/tui/commands.go
+++ b/cli/tui/commands.go
@@ -1760,8 +1760,15 @@ func cmdVPCPrefixDelete(s *Session, args []string) error {
 }
 
 func cmdTenantAccountList(s *Session, _ []string) error {
-	LogCmd(s, "tenant-account", "list")
-	items, err := s.Resolver.Fetch(context.Background(), "tenant-account")
+	ctx := context.Background()
+	var extraFlags []string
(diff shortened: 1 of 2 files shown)


parmani-nv (approved):

thossain-nv (approved):
Adding a note that we'll drop the requirement for specifying this ID in future, but we should merge this for now.

hwadekar-nv (approved):

PR #414, opened 2026-04-22
fix: Mark ipBlockId as required in VPC Prefix create request

kfelternv:
IP Block id is a required argument when creating a VPC Prefix, update the openapi spec and CLI/TUI to handle it correctly.

```
	// NOTE: model validation ensures non-nil IPv4BlockID
	ipBlock, err := common.GetIPBlockFromIDString(ctx, nil, *apiRequest.IPBlockID, csh.dbSession)
```


hwadekar-nv (approved):
LGTM

PR #430, opened 2026-04-27
fix: Create VPC in 'provisioning' state

bcavnvidia:
## Description
VPCs are created with synchronous calls, but the VNI isn't know until it's returned by the site when not sent by the caller.  This means the VPC could be reported as ready before -rest knows the VNI, and then it would suddenly report the VNI without a status change in the VPC.
We now start VPC creation in a "provisioning" state.  We then take the response from the temporal call and pull the VNI from there and make a best-effort attempt to update the DB with the new details.   if all goes well, we return the updated details.  If not, we let inventory processes populate the VNI and update the status, which they already do.

## Type of Change

## Services Affected

## Breaking Changes

## Testing

### DIFF
--- a/api/pkg/api/handler/vpc.go
+++ b/api/pkg/api/handler/vpc.go
@@ -336,7 +336,7 @@ func (cvh CreateVPCHandler) Handle(c echo.Context) error {
 		RoutingProfile:            routingProfile,
 		NVLinkLogicalPartitionID:  defaultNvllPartitionId,
 		Labels:                    labels,
-		Status:                    cdbm.VpcStatusReady,
+		Status:                    cdbm.VpcStatusProvisioning,
 		CreatedBy:                 *dbUser,
 		Vni:                       apiRequest.Vni,
 	}
@@ -362,8 +362,8 @@ func (cvh CreateVPCHandler) Handle(c echo.Context) error {
 
 	// Create status detail
 	sdDAO := cdbm.NewStatusDetailDAO(cvh.dbSession)
-	ssd, err := sdDAO.CreateFromParams(ctx, tx, vpc.ID.String(), *cdb.GetStrPtr(cdbm.VpcStatusReady),
-		cdb.GetStrPtr("VPC successfully provisioned on Site"))
+	ssd, err := sdDAO.CreateFromParams(ctx, tx, vpc.ID.String(), cdbm.VpcStatusProvisioning,
+		cdb.GetStrPtr("VPC provisioning has been initiated on Site"))
 	if err != nil {
 		logger.Error().Err(err).Msg("error creating Status Detail DB entry")
 		return cutil.NewAPIErrorResponse(c, http.StatusInternalServerError, "Failed to create Status Detail for VPC", nil)
@@ -446,7 +446,9 @@ func (cvh CreateVPCHandler) Handle(c echo.Context) error {
 	logger.Info().Str("Workflow ID", wid).Msg("executed synchronous create VPC workflow")
 
 	// Block until the workflow has completed and returned success/error.
-	err = we.Get(ctx, nil)
+	controllerVpc := &cwssaws.Vpc{}
+
+	err = we.Get(ctx, controllerVpc)
 	if err != nil {
 		var timeoutErr *tp.TimeoutError
 		if errors.As(err, &timeoutErr) || err == context.DeadlineExceeded || ctx.Err() != nil {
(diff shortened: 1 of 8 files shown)


workflow/pkg/activity/vpc/vpc.go
```diff
@@ -516,7 +516,7 @@ func (mv ManageVpc) UpdateVpcsInDB(ctx context.Context, siteID uuid.UUID, vpcInv
 
 		// If VPC is not in Deleting state, then update status to Ready
 		if vpc.Status != cdbm.VpcStatusDeleting && vpc.Status != cdbm.VpcStatusReady {
-			err = mv.updateVpcStatusInDB(ctx, nil, vpc.ID, cdb.GetStrPtr(cdbm.VpcStatusReady), cdb.GetStrPtr("VPC has been re-detected on Site"))
+			err = mv.updateVpcStatusInDB(ctx, nil, vpc.ID, cdb.GetStrPtr(cdbm.VpcStatusReady), cdb.GetStrPtr("VPC has been detected on Site"))
```
thossain-nv: The message in the status detail was covering an obscure case where VPC was mising on Site and then re-appeared. Since we're using it for a proper readiness case now, we should probably say:
```
VPC is ready for use
```

thossain-nv (commented):
Thanks @bcavnvidia, some thought.

1. We could make a best effort pull from Site like we do for DPU Extension service 
2. We could also update VPC creation Site Workflow to return the VPC object and capture it in `err = we.Get(ctx, nil)` (instead of `nil`). As long as Site Agent is deployed ahead, this will be a seamless transition. This would be preferable and we did this for another object recently. 

Inventory update can always be a fallback.

Since VPC creation is immediate on Site, making users wait for inventory may cause user frustration so either of the above

workflow/pkg/activity/vpc/vpc.go
bcavnvidia: > We could also update VPC creation Site Workflow to return the VPC object and capture it in

I wanted that at first, but it doesn't quite get around the changes here.

You then have to make another DB call to update the DB VNI.  If you do that within the same txn, and it fails for some reason, you risk orphaning the -core VPC because inventory doesn't store VPCs that it sees only existing in -core.  If you do it in a _different_ txn and it fails, then you'd return a failure to the caller that could be confusing because the VPC was created and stored, but the call is returning a failure, and then you still need to wait for inventory.

> We could make a best effort pull from Site like we do for DPU Extension service

Similar to the other one, if it's critical path, you'd have the same failure cases.  If it's best-effort, you'd still need to mark the VPC as non-ready if your best-effort fails so that the state can move to ready after an inventory pull.

So, TL;DR:  we could best-effort the VNI update so that a caller doesn't have to wait/poll for the status to change under happy-path circumstances, but it doesn't seem to let you avoid the overall changes here.  It'd be in addition to them.

bcavnvidia: Re-reading your comment, it looks like you are advocating for grabbing from the workflow response but then allowing the DB update to fail silently (logging only)

thossain-nv: Right, if we only log the post-creation update failures, then we gain the instant update and in case of failure, let inventory catch up to it.

bcavnvidia: And in that case, we shouldn't return the VNI in the response if the DB  update were to fail for some reason; otherwise, users would see it
 appearing, disappearing, and then reappearing again.

hwadekar-nv: Just a thought: if we receive `controllerVpc.Status.Vni` in the inventory, should we consider the VPC ready?

api/pkg/api/handler/vpc.go
```diff
@@ -495,8 +497,56 @@ func (cvh CreateVPCHandler) Handle(c echo.Context) error {
...
+				Status:    cdb.GetStrPtr(cdbm.VpcStatusReady),
+			}
+			updatedVpc, err := vpcDAO.Update(ctx, tx, uvpcInput)
+			if err != nil {
+				logger.Error().Err(err).Msg("error while updating VPC DB entry for VNI")
+				rollback = true
+			} else {
+				// Create status detail
+				sdDAO := cdbm.NewStatusDetailDAO(cvh.dbSession)
+				ssd, err = sdDAO.CreateFromParams(ctx, tx, vpc.ID.String(), cdbm.VpcStatusReady, cdb.GetStrPtr("VPC is ready for use"))
```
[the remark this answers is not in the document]
bcavnvidia: @coderabbitai  VNI is currently assigned immediately.  If GetStatus() is not nil, VNI will not be nil.  If VPC creation in the backend service changes to be a longer cycle, it will require that it return an explicit value for determining readiness, and that would then be applied/tested here.

workflow/pkg/activity/vpc/vpc.go
bcavnvidia: Oh, didn't see this comment.  We should, yes.  I think I told coderabbit that as well.  At the moment, even controllerVpc.GetStatus() != nil is directly tied to it being in a ready state.

bcavnvidia: The VPC is created immediately as always at the moment.  It's just that we now need to grab and store the VNI that the site assigns to it for the case of auto-allocated VNIs (the original/common case).

api/pkg/api/handler/vpc.go
```diff
@@ -484,8 +486,58 @@ func (cvh CreateVPCHandler) Handle(c echo.Context) error {
...
 	txCommitted = true
+
+	statusDetails := []cdbm.StatusDetail{*ssd}
+
+	// Make a best-effort attempt to return a response with the allocated VNI.
+	if controllerVpc.GetStatus() != nil {
+		activeVni := wutil.GetUint32PtrToIntPtr(controllerVpc.GetStatus().Vni)
+
+		// Start a database transaction
+		tx, err := cdb.BeginTx(ctx, cvh.dbSession, &sql.TxOptions{})
```
thossain-nv: For best effort DB calls, we can omit the transaction. I'd take any call that goes through. Keeps this code section short too.

thossain-nv (approved):
Thanks @bcavnvidia, one last suggestion.

api/pkg/api/handler/vpc.go
bcavnvidia: Ok, as long as we're comfortable with a potentially missing status detail even when the status of the VPC successfully changed, we can toss the txn.

PR #442, opened 2026-04-29
feat: Support fw updates to unregistered devices in NSM and PSM

spydaNVIDIA:
## Description

feat: Support fw updates to unregistered devices in NSM and PSM

## Type of Change

## Services Affected

## Breaking Changes

## Testing


nvswitch-manager/internal/service/server_impl.go
```diff
@@ -426,22 +448,41 @@ func (s *NVSwitchManagerServerImpl) QueueUpdates(ctx context.Context, req *pb.Qu
...
-		if err != nil {
-			result.Status = pb.StatusCode_INTERNAL_ERROR
-			result.Error = fmt.Sprintf("failed to queue update: %v", err)
+		s.queueFirmwareUpdate(ctx, result, switchUUID, req.BundleVersion, components)
+		results = append(results, result)
+	}
+
+	// Direct path: auto-register unregistered targets, then queue updates
+	for _, target := range req.Targets {
+		result := &pb.QueueUpdateResult{
```
kunzhao-nv, line 481: Missing null check for `target`

nvswitch-manager/pkg/credentials/inmemory.go
```diff
@@ -83,11 +83,19 @@ func (m *InMemoryCredentialManager) GetBMC(ctx context.Context, mac net.Hardware
...
 	defer m.mu.Unlock()
 	key := m.bmcKey(mac)
+	if existing, exists := m.store[key]; exists {
+		if existing.Equal(cred) {
+			log.Infof("BMC credentials for %s already exist and match; skipping write", mac)
+			return nil
+		}
+		log.Warnf("BMC credentials for %s differ from existing; overwriting in-memory entry", mac)
+	}
 	m.store[key] = cred
```
kunzhao-nv, line 104: What if cred is null here?

powershelf-manager/internal/service/server_impl.go
```diff
@@ -190,25 +200,67 @@ func (s *PowershelfManagerServerImpl) listAvailableFirmware(ctx context.Context,
...
 
+	// Direct path: auto-register unregistered targets, then upgrade
+	for _, targetReq := range req.Targets {
+		target := targetReq.Target
+		resp := &pb.UpdatePowershelfFirmwareResponse{
+			PmcIp: target.GetPmcIpAddress(),
+		}
+
+		if err := validateFirmwareTarget(target); err != nil {
+			resp.Components = []*pb.UpdateComponentFirmwareResponse{{
```
kunzhao-nv: When a target is requested with N components, the success path returns N UpdateComponentFirmwareResponse entries via upgradeComponents, but both the validate-failure branch (220-227) and the registration-failure branch (229-237) emit a single one. Callers that correlate by index can't handle the shape mismatch. Please loop over targetReq.Components and emit one entry per requested component with the same Status/Error, mirroring the success path.

powershelf-manager/internal/service/server_impl.go
```diff
@@ -190,25 +200,67 @@ func (s *PowershelfManagerServerImpl) listAvailableFirmware(ctx context.Context,
...
+
 	return &pb.UpdateFirmwareResponse{
 		Responses: responses,
 	}, nil
 }
 
+// upgradeComponents upgrades each requested component for a given PMC MAC and returns per-component results.
+func (s *PowershelfManagerServerImpl) upgradeComponents(ctx context.Context, pmcMac string, components []*pb.UpdateComponentFirmwareRequest) []*pb.UpdateComponentFirmwareResponse {
+	results := make([]*pb.UpdateComponentFirmwareResponse, 0, len(components))
+	for _, component := range components {
```
kunzhao-nv, line 296: UpdateComponentFirmwareResponse.Component is never populated — neither by the UpgradeTo == nil branch in upgradeComponents nor by updateFirmware's success/error returns. PMC was the only component before, so this stayed hidden, but now that the proto allows multi-component requests via FirmwareTarget callers can't correlate status back to the requested component. Set Component: component.Component on the validate path and resp.Component = pbComponent inside updateFirmware.

nvswitch-manager/internal/service/service.go
```diff
@@ -50,20 +50,17 @@ type Service struct {
 func New(ctx context.Context, c Config) (*Service, error) {
 	// Connect to database first if configured (needed for persistent registry)
 	var db *bun.DB
-	if c.DBConf.Host != "" {
+	if c.DataStoreType == nvswitchmanager.DatastoreTypePersistent && c.DBConf.Host != "" {
```
kunzhao-nv: The && c.DBConf.Host != "" guard means a misconfigured persistent-mode setup (DataStoreType=persistent, Host empty) silently skips DB init and leaves db == nil, deferring the failure to runtime instead of fail-fast at startup. Suggest:

```
if c.DataStoreType == nvswitchmanager.DatastoreTypePersistent {
    if c.DBConf.Host == "" {
        return nil, fmt.Errorf("DB host is required for persistent mode")
    }
    ...
}

File: nvswitch-manager/pkg/credentials/vault.go [hunk 6]
```diff
@@ -195,19 +195,57 @@ func (m *VaultCredentialManager) delete(ctx context.Context, key string) error {
...
+
 // GetBMC retrieves and validates BMC credentials for the given MAC from Vault.
 func (m *VaultCredentialManager) GetBMC(ctx context.Context, mac net.HardwareAddr) (*credential.Credential, error) {
 	return m.get(ctx, m.getBMCCredentialKey(mac))
 }
 
-// PutBMC writes the BMC credentials to Vault.
+// PutBMC writes BMC credentials to Vault. If an entry already exists and matches,
+// this is a no-op. If an entry exists but differs, an error is returned (use PatchBMC to overwrite).
 func (m *VaultCredentialManager) PutBMC(ctx context.Context, mac net.HardwareAddr, cred *credential.Credential) error {
```
kunzhao-nv, line 268: Two potential issues with the new conditional-write semantics:

1. m.get(...) returns errors for four reasons (vault read failure, secret missing, unexpected data shape, parse failure). Only "secret missing" is a legitimate fall-through to m.put. As written, a transient Vault hiccup or a malformed existing secret followed by a Put with new credentials silently overwrites the stored value. Introduce a typed sentinel for the not-found case and gate the write on errors.Is(err, errCredentialNotFound).
2. The new "differ → error" contract diverges from the in-memory backend (still warn-and-overwrite). All Register callers go through Put* unconditionally (nvswitchmanager.Register at lines 95/101, pmcmanager.Register at line 73). If persistent mode is ever enabled, RLA inventory sync and PSM register tasks will fail on credential rotation. Please align both backends — either both reject conflicts (and migrate every register caller to Patch*), or both warn-and-overwrite. Same applies to PutNVOS (line 263) and PSM's Put (powershelf-manager/pkg/credentials/vault.go:174).

nvswitch-manager/internal/service/server_impl_test.go
```diff
@@ -95,3 +96,128 @@ func TestResetTarget_EmptyCredentials(t *testing.T) {
...
+	// by checking that the handler returns results (not a gRPC error) for invalid targets.
+
+	// Without fwm, QueueUpdates returns "firmware manager not initialized"
+	// So we only test validateFirmwareTarget directly here.
+	target := validFirmwareTarget()
+	target.BmcIp = "invalid"
+	err := validateFirmwareTarget(target)
+	assert.Error(t, err)
+
+	_ = srv // suppress unused
```
kunzhao-nv: nit: TestQueueUpdates_InvalidTarget constructs srv but never uses it (_ = srv to suppress unused). The body just re-tests validateFirmwareTarget, which is already covered by TestValidateFirmwareTarget_InvalidBmcIP. Either drive srv.QueueUpdates(...) end-to-end (e.g. assert it returns the expected Unavailable when fwm is nil) or delete this test.

nvswitch-manager/pkg/credentials/vault_test.go
```diff
@@ -349,6 +349,54 @@ func TestVaultManager_NVOSPutGet(t *testing.T) {
...
+	// First put succeeds
+	assert.NoError(t, mgr.PutBMC(ctx, mac, newCredential("admin", "secret")))
+	assert.NoError(t, mgr.PutNVOS(ctx, mac, newCredential("nvos", "nvos_secret")))
+
+	// Idempotent put with same credentials succeeds (no-op)
+	assert.NoError(t, mgr.PutBMC(ctx, mac, newCredential("admin", "secret")))
+	assert.NoError(t, mgr.PutNVOS(ctx, mac, newCredential("nvos", "nvos_secret")))
+
+	// Put with different credentials returns error
+	assert.Error(t, mgr.PutBMC(ctx, mac, newCredential("admin", "different")))
```
kunzhao-nv: assert.Error here passes for any Vault failure including transient/network errors, so it doesn't actually pin the new "conflict" contract. Either assert.Contains(t, err.Error(), "differ from provided") or expose a sentinel error and use errors.Is. Same in PSM's matching test.

nvswitch-manager/pkg/credentials/inmemory_test.go
```diff
@@ -104,15 +109,15 @@ func TestInMemoryBMCPutGet(t *testing.T) {
...
-				// For the overwrite scenario, put a second credential to same MAC
-				if name == "put overwrites existing value" {
-					assert.NoError(t, mgr.PutBMC(ctx, mac, credential.New("user2", "p2")))
-				}
+		// Optional initial put
+		if tc.initialPut {
+			mac := parseMAC(t, tc.putMAC)
+			assert.NoError(t, mgr.PutBMC(ctx, mac, tc.putCred))
+			// For the idempotent scenario, put the same credential again
+			if name == "put same credential is no-op" {
```
kunzhao-nv: nit:
Branching on name == "put same credential is no-op" means renaming the case silently disables the idempotent path. Promote it to a boolean field on the table case (e.g. putSameAgain bool) and key the second PutBMC off that.

nvswitch-manager/internal/proto/v1/nvswitch-manager.proto
```diff
@@ -263,10 +277,12 @@ message QueueUpdateResponse {
 }
 
 // QueueUpdatesRequest queues firmware updates for multiple switches.
+// Registered switches are identified by UUID; unregistered devices use FirmwareTarget.
 message QueueUpdatesRequest {
```
kunzhao-nv, line 297: I am thinking if we can better design this. A few questions:
1. It is valid to fill both switch_uuids and targets? (seem so for current impl but should it)
2. What if a device, appear in both witch_uuids and targets? There is not dedup logic so we will update firmware twice?
3. nit: targets is best effort now so we'd better add comments on top
4. components is also here. If user want to update A.BMC and B.NVOS, user can not bind them in a single call now (it is OK if this is intended)

nvswitch-manager/pkg/credentials/inmemory.go
spydaNVIDIA: added check

powershelf-manager/internal/service/server_impl.go
spydaNVIDIA: fixed

nvswitch-manager/pkg/credentials/vault.go
spydaNVIDIA: fixed: I modified persistent mode to match in memory

nvswitch-manager/internal/proto/v1/nvswitch-manager.proto
spydaNVIDIA: I addressed 1 & 2 by making it invalid to specify both switch_uuids and targets. that validation is enforced in the handler.

Addressed 3 by updating the comment. Both targets and switch uuid are best effort right.

4 is intended and was the behavior even before this MR. if we want to address this, we can fix it elsewhere.

powershelf-manager/internal/service/server_impl.go
```diff
@@ -190,61 +202,162 @@ func (s *PowershelfManagerServerImpl) listAvailableFirmware(ctx context.Context,
...
-			PmcMacAddress: pmc_mac,
-			Components:    componentUpgradeResponses,
+			PmcMacAddress: powershelf.PmcMacAddress,
+			Components:    s.upgradeComponents(ctx, powershelf.PmcMacAddress, powershelf.Components),
 		})
 	}
 
+	// Direct path: auto-register unregistered targets, then upgrade
+	for _, targetReq := range req.Targets {
+		target := targetReq.Target
```
kunzhao-nv, line 241: Null check for `targetReq`

kunzhao-nv (approved):
Need to update `inventorysync` in RLA accordingly.

PR #449, opened 2026-04-30
chore: Reuse pre-fetched Site in Expected Component GET handlers

chet:
## Description

Continuing the cleanup across the `ExpectedMachine`, `ExpectedPowerShelf`, and `ExpectedSwitch` handlers. Each `Get` handler currently does an **extra** database call: it loads the entity with the caller's `qIncludeRelations`, and then issues a separate `siteDAO.GetByID` purely to satisfy the access check. When the caller already passed `?includeRelation=Site`, it turns into two queries for the same `Site` row.

`Update` and `Delete` already load `Site` as part of the entity `Get` and use `entity.Site` directly. This just teaches `Get` to do the same when it has the relation, falling back to the separate fetch otherwise. API behavior remains the same -- the response only includes `Site` when the caller asked for it.

**Note! We *could* go further and *always* join `Site`, dropping `siteDAO.GetByID` entirely. That'd match how `Update` and `Delete` *already* work, but `Get` responses would then always include `Site` whether the caller asked for it or not. Just a thought.**

## Type of Change

## Services Affected

## Related Issues (Optional)

None

## Breaking Changes

## Testing

- [x] No testing required (docs, internal refactor, etc.)

## Additional Notes

None

### DIFF
--- a/api/pkg/api/handler/expectedmachine.go
+++ b/api/pkg/api/handler/expectedmachine.go
@@ -596,12 +596,15 @@ func (gemh GetExpectedMachineHandler) Handle(c echo.Context) error {
 		return cutil.NewAPIErrorResponse(c, http.StatusInternalServerError, "Failed to retrieve Expected Machine due to DB error", nil)
 	}
 
-	// Get Site for the Expected Machine
-	siteDAO := cdbm.NewSiteDAO(gemh.dbSession)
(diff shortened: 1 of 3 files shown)


thossain-nv (approved):
Looks good, thanks @chet

PR #465, opened 2026-05-01
fix: Add API routes for Expected Machine batch handlers

kfelternv:
## Description

Currently Expected Machine batch API handlers are not registered, this is causing CLI commands to fail. This PR adds the missing route entries.

## Type of Change

## Services Affected

## Related Issues (Optional)

None

## Breaking Changes

## Testing

## Additional Notes

None

### DIFF
--- a/api/pkg/api/routes.go
+++ b/api/pkg/api/routes.go
@@ -427,6 +427,16 @@ func NewAPIRoutes(dbSession *cdb.Session, tc tClient.Client, tnc tClient.Namespa
 			Handler: apiHandler.NewGetAllExpectedMachineHandler(dbSession, cfg),
 		},
 		{
+			Path:    apiPathPrefix + "/expected-machine/batch",
+			Method:  http.MethodPost,
+			Handler: apiHandler.NewCreateExpectedMachinesHandler(dbSession, scp, cfg),
+		},
+		{
+			Path:    apiPathPrefix + "/expected-machine/batch",
+			Method:  http.MethodPatch,
+			Handler: apiHandler.NewUpdateExpectedMachinesHandler(dbSession, scp, cfg),
+		},
+		{
 			Path:    apiPathPrefix + "/expected-machine/:id",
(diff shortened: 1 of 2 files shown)


api/pkg/api/handler/expectedmachine_test.go
```diff
@@ -1045,6 +1045,20 @@ func TestUpdateExpectedMachineHandler_Handle(t *testing.T) {
 				assert.Equal(t, "AA:BB:CC:DD:EE:FF", response.BmcMacAddress, "MAC address in response should match the updated value")
 			},
 		},
+		{
+			name:        "no update fields should return 400",
```
pbreton: Is that the general pattern in handlers?

kfelternv: Not sure, this will resolve the silent no update scenario. Is there another way to solve it that is already a pattern in the codebase? Or should we leave it as is?

hwadekar-nv: In case of Nvlink Interface no-op update, we are thinking to return existing object as it is

kfelternv: The body is pretty clear as to why it was 400 error, `"no updates specified"`

I think this is ok as is.

thossain-nv: The general pattern we currently we follow for REST API updates is that if nothing changed, the existing object is returned. So it becomes an idempotent call.

The same pattern is in Core, if nothing has changed, the update is still accepted.

Do we want to change this behavior?

kfelternv: That sounds good, sticking to the existing pattern seems like the right call. I wasn't sure if it was an established pattern or if that only applied to the NVlink interface. I'll update this to match the pattern.

thossain-nv (approved):
Looks good, thanks @kfelternv

PR #470, opened 2026-05-02
feat: Allow updating Site capabilities using Site update API endpoint

thossain-nv:
## Description

Allows Provider to enable or disable supported Site capabilities through Site update API endpoint.

## Type of Change

## Services Affected

## Related Issues (Optional)

Addresses: https://github.com/NVIDIA/infra-controller-rest/issues/425

## Breaking Changes

## Testing

## Additional Notes

None


api/pkg/api/handler/site.go
```diff
@@ -495,6 +495,16 @@ func (ush UpdateSiteHandler) Handle(c echo.Context) error {
 			}
 		}
 
+		if apiRequest.Capabilities != nil {
+			siteUpdateInput.Config = &cdbm.SiteConfigUpdateInput{
+				NativeNetworking:          apiRequest.Capabilities.NativeNetworking,
```
bcavnvidia, line 520: If we block a user from switching a VPC to FNN when FNN isn't enabled, do we still allow disabling FNN if there are already FNN VPCs?

Just curious.  Still makes sense to me if this is just a gate against _new_ things being created and we'd just document it as such.  E.g., disabling FNN is a way to prevent _new_ FNN VPCs for whatever reason the provider has.

db/pkg/db/model/site.go
```diff
@@ -593,10 +593,9 @@ func (ssd SiteSQLDAO) Update(ctx context.Context, tx *db.Tx, input SiteUpdateInp
 	}
 
 	if input.Config != nil {
-
 		_, err := db.GetIDB(tx, ssd.dbSession).NewUpdate().
 			Model(st).
-			Set("config = config || ?::jsonb", input.Config).
+			Set("config = config || ?::jsonb, updated = current_timestamp", input.Config).
```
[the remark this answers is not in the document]
bcavnvidia: @coderabbitai In PostgreSQL, if you add a new column with a DEFAULT clause, existing rows will automatically get that default value.  A 'capabilities' column was added with a default value.  Later, a 'config' column was added with a default value and existing values in 'capabilities' were migrated to 'config'.  Review migrations for the two columns to confirm.

api/pkg/api/handler/site.go
thossain-nv: We can block them from disabling Native Networking if they have VPC Prefixes or FNN VPCs. Maybe we ask users that they delete those before disabling the flag?

sdk/standard/model_site_capabilities_update_request.go
```diff
@@ -0,0 +1,328 @@
+/*
+NVIDIA Infra Controller REST API
+
+NVIDIA Infra Controller REST API allows users to create and manage resources e.g. VPC, Subnets, Instances across all connected NVIDIA Infra Controller datacenters, also referred to as Sites.
+
+API version: 1.4.0
+*/
+
+// Code generated by OpenAPI Generator (https://openapi-generator.tech); DO NOT EDIT.
+
```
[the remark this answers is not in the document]
kfelternv: You should be able to update with the latest main, then regenerate the sdk files. I've added a script that should automatically add these headers when they are generated using the make command.

kfelternv:
> You should be able to update with the latest main, then regenerate the sdk files. I've added a script that should automatically add these headers when they are generated using the make command.

@thossain-nv I think this should knock down the diff significantly for this PR.

api/pkg/api/handler/site.go
hwadekar-nv: Currently we only allow them ETHER based to FNN but not other

hwadekar-nv (approved):

thossain-nv:
~@kfelternv Is the comment you quoted from another PR?~ Nevermind I see what you're saying, just need to rebase.

api/pkg/api/handler/site.go
```diff
@@ -495,6 +495,16 @@ func (ush UpdateSiteHandler) Handle(c echo.Context) error {
...
+				NetworkSecurityGroup:      apiRequest.Capabilities.NetworkSecurityGroup,
```
nvlitagaki, line 521: In a similar vein to the previous comments, we should probably prevent the following and ask them to delete the relevant records first

- disabling NSG if there are already NSGs on the site
- disabling NVLink if there are already NVLink partitions
- disabling RLA if there are already racks, trays, expected switches, expected power shelves or expected racks (though that last one depends on [this PR](https://github.com/NVIDIA/infra-controller-rest/pull/444))
- disabling image based operating systems if there are image-based OSes on the site

thossain-nv: Disabling these flags only affect create/update operations, still allows deleting existing resource. We should allow the Provider to modify the flags to temporarily or permanently disable creation/update of these resources if they want.

For example: disabling network security group flag only prevents creation/update. Existing ones can still be used to create Instances. Similarly disabling image based flag prevents using image based OS creation or using them in for new Instances.

Disabling them and continuing with regular operations won't cause any issues, except for FNN - where we rely on VPC's virtualization type to make some decisions instead of checking Site capability. So we treat that as a special case.

In general I think we should allow users to make these changes even if there are existing objects. Most Providers will likely use this to enable certain flags.

PR #462, opened 2026-04-30
refactor: Introduce db WithTx + migrate Expected handlers

chet:
## Description

Following on all of the `Expected*` cleanup work I've been doing, I was looking at the database bits, and wanted to attempt to clean things up a bit even more there.

This is modeled on the `with_txn` pattern from the "Core" `api-db` crate, and adds four helpers:
```
      cdb.WithTx(ctx, dbSession, func(tx *cdb.Tx) error { ... })
      cdb.WithTxOpts(ctx, dbSession, opts, fn)              // to provide non-default sql.TxOptions
      cdb.WithTxResult(ctx, dbSession, fn)                  // where the closure returns (T, error)
      cdb.WithTxResultOpts(ctx, dbSession, opts, fn)
```

Where if the closure:
- Returns `nil`, the `tx` is committed.
- Returns an `err`, the `tx`is rolled back.

Callers no longer manage `Begin`/`Commit`/`Rollback` or maintain a "did we commit?" flag manually -- the closure boundary IS the commit boundary. The whole class of "*forgot to set txCommitted = true after Commit()*" bugs goes away because there's no flag to forget.

Example usage:
```
  err := cdb.WithTx(ctx, dbSession, func(tx *cdb.Tx) error {
      if err := helloDAO.Update(ctx, tx, input); err != nil {
          return cutil.NewAPIError(http.StatusInternalServerError, "DB update failed", nil)
      }
      return nil
  })
  if err != nil {
      return common.HandleTxError(c, logger, err, "Failed to update hello, txn rolled back")
  }
```
For now, I'm only wiring up the `Expected*` handlers to keep the size of this PR small, so it's easier to understand what's going on, what things look like after being refactored, etc.

Assuming we like it, I can do another PR that pulls more handlers in.

Tests added!

Signed-off-by: Chet Nichols III <chetn@nvidia.com>

## Type of Change

- [ ] **Feature** - New feature or functionality (feat:)
- [ ] **Fix** - Bug fixes (fix:)
- [ ] **Chore** - Modification or removal of existing functionality (chore:)
- [x] **Refactor** - Refactoring of existing functionality (refactor:)
- [ ] **Docs** - Changes in documentation or OpenAPI schema (docs:)
- [ ] **CI** - Changes in GitHub workflows. Requires additional scrutiny (ci:)
- [ ] **Version** - Issuing a new release version (version:)

## Services Affected

- [x] **API** - API models or endpoints updated
- [ ] **Workflow** - Workflow service updated
[...]

PR #472, opened 2026-05-02
refactor: Migrate vpc handlers to WithTx

chet:
## Description

This applies the new `WithTx` pattern from #462 to the `vpc` handlers (Create/Update/Delete plus the virtualization-type update).

The `UpdateVirtualization` handler uses the `timeoutResp` closure pattern for `common.TerminateWorkflowOnTimeOut` so the helper still runs after the tx unwinds; the other three retain their inline `stc.TerminateWorkflow` paths.

Keeping these PRs smaller and tightly scoped so they're:
- In theory a little easier to read.
- More tightly scoped/less "blast radius" per PR.
- A little nicer on/for @coderabbitai. 😆

I do I wish the diffs were easier to read, but it is what it is!

## Type of Change

- [x] **Refactor** - Refactoring of existing functionality (refactor:)

## Services Affected

## Breaking Changes

## Testing

### DIFF
--- a/api/pkg/api/handler/vpc.go
+++ b/api/pkg/api/handler/vpc.go
@@ -19,7 +19,6 @@ package handler
 
 import (
 	"context"
-	"database/sql"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


api/pkg/api/handler/vpc.go
```diff
@@ -312,180 +311,172 @@ func (cvh CreateVPCHandler) Handle(c echo.Context) error {
...
+				}
 
-		code, err := common.UnwrapWorkflowError(err)
-		logger.Error().Err(err).Msg("failed to synchronously execute Temporal workflow to create VPC")
-		return cutil.NewAPIErrorResponse(c, code, fmt.Sprintf("Failed to execute sync workflow to create VPC on Site: %s", err), nil)
-	}
+				logger.Info().Str("Workflow ID", wid).Msg("initiated terminate synchronous create VPC workflow successfully")
 
-	logger.Info().Str("Workflow ID", wid).Msg("completed synchronous create VPC workflow")
+				return cutil.NewAPIError(http.StatusInternalServerError, fmt.Sprintf("Failed to create VPC, timeout occurred executing workflow on Site: %s", wferr), nil)
```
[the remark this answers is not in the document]
chet: @coderabbitai Ahhh yes -- this is also primarily another case of pre-existing logic made more visible from doing this refactoring, BUT, totally good call out. Let me fix. Thanks!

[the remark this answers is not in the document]
chet: @coderabbitai I think we're good here? Just pushed up the tweak. Let me know. Thanks!

api/pkg/api/handler/vpc.go
```diff
@@ -312,180 +311,182 @@ func (cvh CreateVPCHandler) Handle(c echo.Context) error {
...
-	if err != nil {
-		var timeoutErr *tp.TimeoutError
-		if errors.As(err, &timeoutErr) || err == context.DeadlineExceeded || ctx.Err() != nil {
+		// Block until the workflow has completed and returned success/error.
+		wferr = we.Get(wfCtx, controllerVpc)
+		if wferr != nil {
+			var timeoutErr *tp.TimeoutError
+			if errors.As(wferr, &timeoutErr) || wferr == context.DeadlineExceeded || wfCtx.Err() != nil {
+				logger.Error().Err(wferr).Msg("failed to create VPC, timeout occurred executing workflow on Site.")
+				timeoutResp = func() error {
```
hwadekar-nv, line 458: I’m trying to understand how this cancel workflow will execute based on this pattern.

thossain-nv, line 458: Should we make use if the existing utility method [TerminateWorkflowOnTimeOut](https://github.com/NVIDIA/infra-controller-rest/blob/main/api/pkg/api/handler/util/common/common.go#L1093) for the timeout handling?

thossain-nv (approved):
Thank you for the great refactor @chet Added a suggestion for timeout handling.

api/pkg/api/handler/vpc.go
thossain-nv: WithTx returns APIError, which gets processed after WithTx executes. Other HandleTxError manages returning error.

chet: @hwadekar-nv  Yeahhh good question! The `timeoutResp` flow is:

1. Inside the closure, when we detect a timeout, we don't call `TerminateWorkflow` directly. Instead, we stash a closure in the outer-scope `timeoutResp` that knows how to do the termination work.
2. We then return a `*cutil.APIError` from the `WithTx` closure, which causes `WithTx` to roll back the transaction.
3. After `WithTx` returns (rollback complete, locks released, etc), the outer code checks if `timeoutResp` is set; that's when the actual `TerminateWorkflow` RPC fires (calling `TerminateWorkflowOnTimeOut`, which builds a fresh context with `WorkflowContextNewAfterTimeout` + hits Site to terminate the orphaned workflow + and returns an error to the caller.

And if it's not obvious at first glance, the reason we don't inline `TeriminateWorkflow` while still inside the closure is because the DB transaction would stay _open_ during that second RPC, which would continue to hold locks + a connection while we wait on the network/RPC.

We ran into this a bunch in the Core codebase, and have spent almost year trying to clean it up and unwind it (we even have a custom compiler extension now that looks for blocking calls during transactions).

Lemme know if that all makes sense!

chet: @thossain-nv omg yes! amazing.

chet: done!

PR #480, opened 2026-05-04
feat: Support IP Usage stats in VPC Prefix and Subnet

hwadekar-nv:
## Description

Here is the strategy we are considering to get prefix usage for VPC Prefix and Subnet.

**VPC Prefix**
- AvailableIPs = Total Number of IPs from prefix - 2 * number of interfaces
- AvailablePrefixes = First 10 available /31 prefixes (total count = Available IPs/2)
- AcquiredIPs = Number of interfaces * 2
- AcquiredPrefixes = Number of interfaces

**Subnet**
- AvailableIPs = Total Number of available IPs from prefix -  number of interfaces - 2
- AvailablePrefixes = 0, we consider prefixes to be at least `/31`
- AcquiredIPs = Number of Interfaces + 2
- AcquiredPrefixes = 0

## Type of Change

## Services Affected

## Related Issues (Optional)

None

## Breaking Changes

## Testing

## Additional Notes

None


api/pkg/api/handler/subnet.go
```diff
@@ -569,12 +588,30 @@ func (gash GetAllSubnetHandler) Handle(c echo.Context) error {
...
+				continue
+			}
+			prefixUsage, serr := ipam.GetIpamUsageForSubnet(ctx, ipamStorage, sn, sn.IPv4Block)
+			if serr != nil {
+				logger.Error().Err(serr).Str("subnetId", sn.ID.String()).Msg("error retrieving ipam usage stats for Subnet")
+				continue
+			}
+			puSubnetMap[sn.ID] = prefixUsage
+		}
+	}
```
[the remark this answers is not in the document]
hwadekar-nv: Here, `IPv4Block` wouldn’t be `nil`, so it’s okay to continue in case it’s accidentally deleted by manually from the database.

api/pkg/api/handler/vpcprefix.go
```diff
@@ -500,13 +508,31 @@ func (gash GetAllVpcPrefixHandler) Handle(c echo.Context) error {
...
+				continue
+			}
+			prefixUsage, serr := ipam.GetIpamUsageForVpcPrefix(ctx, ipamStorage, vp, vp.IPBlock)
+			if serr != nil {
+				logger.Error().Err(serr).Msg("error retrieving ipam usage stats for VPC prefix")
+				continue
+			}
+			puVpMap[vp.ID] = prefixUsage
+		}
+	}
```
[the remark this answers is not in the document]
hwadekar-nv: `IPBlock` might not be 'nil' always, hence okey to ignore this

thossain-nv:
@hwadekar-nv I don't think we can use IPAM to get usage stats for Subnets or VPC Prefixes. This is because they get used only when Instances are created and we don't record Instance IP usage in IPAM, that's decided by Core. The IPAM usage stats will not give back any meaningful information.

We have to look into Instance count and Interface count to understand how many IPs have been used and how much IP space is left. We can work with @bcavnvidia to find how much IP space each Interface uses for Subnets and for VPC Prefixes.

openapi/spec.yaml
nvlitagaki, line 1: Please include an example of a VPC Prefix that includes `usageStats`, but be mindful to not accidentally imply that this value can also be included in the return value from a create or update call

hwadekar-nv:
Got it, also IPAM is now configurable too so not always be available, Thanks @thossain-nv, will sync with @bcavnvidia for this.

api/pkg/api/handler/util/common/common.go
```diff
@@ -630,6 +631,23 @@ func HandleTxError(c echo.Context, logger zerolog.Logger, err error, fallback st
 	return cutil.NewAPIErrorResponse(c, http.StatusInternalServerError, fallback, nil)
 }
 
+// ParseIncludeUsageStats reads includeUsageStats from the request; when true, clones relations and appends relationForStats
+// if absent (Subnet: IPv4Block; VPC prefix: IPBlock).
+func ParseIncludeUsageStats(c echo.Context, relations []string, relationForStats string) (includeUsageStats bool, augmented []string, err error) {
```
thossain-nv: Usage stats and `IncludeRelation` are not related. Grouping them in a function does not seem intuitive. Let's leave these in the respective handlers.

hwadekar-nv: Revert the changes

hwadekar-nv:
@bcavnvidia  Would you review this if this looks okey based on Instance and Interface based IP usage?

api/pkg/api/handler/vpcprefix.go
```diff
@@ -618,6 +656,20 @@ func (gsh GetVpcPrefixHandler) Handle(c echo.Context) error {
...
+		if err != nil {
+			return cutil.NewAPIErrorResponse(c, http.StatusBadRequest, "Invalid value specified for `includeUsageStats` query param", nil)
+		}
+	}
+
+	queryIncludeRelations := slices.Clone(qIncludeRelations)
+	if includeUsageStats && !slices.Contains(queryIncludeRelations, cdbm.IPBlockRelationName) {
+		queryIncludeRelations = append(queryIncludeRelations, cdbm.IPBlockRelationName)
+	}
+
```
kfelternv, line 655: I think this logic is duplicated from earlier in the file, might be more efficient to pull this out into a single fuction.

api/pkg/api/model/util/util.go
```diff
@@ -174,3 +182,139 @@ func ProtobufLabelsFromAPILabels(labels map[string]string) []*cwssaws.Label {
 	}
 	return protoLabels
 }
+
+// ipv4UsableHostAddressesForPrefixBits returns assignable IPv4 host count for a prefix length (/32 and /31 special-cased).
+func ipv4UsableHostAddressesForPrefixBits(prefixBits int) uint64 {
+	if prefixBits < 0 || prefixBits > 32 {
```
bcavnvidia: I get the ease of use that returning 0 provides here, but it also feels like returning an error might be right, but 0 is probably fine since it'd probably block any action you were going to take based on the number of available hosts.  It just might mask other issues.

db/pkg/db/ipam/ipam.go
```diff
@@ -145,6 +173,101 @@ func GetIpamUsageForIPBlock(ctx context.Context, ipamDB cipam.Storage, ipBlock *
...
+
+	vpcCidr := getVpcPrefixCidr(ctx, vpcPrefix)
+	parentCidr := GetCidrForIPBlock(ctx, ipBlock.Prefix, ipBlock.PrefixLength)
+
+	if ipBlock.FullGrant && cidrPrefixesEqual(vpcCidr, parentCidr) {
+		return &cipam.Usage{
+			AvailableIPs:              0,
+			AcquiredIPs:               0,
+			AvailableSmallestPrefixes: 0,
+			AvailablePrefixes:         nil,
```
bcavnvidia: What does AvailablePrefixes mean here?  Just the number of prefixes that haven't been totally exhausted?

api/pkg/api/model/util/util.go
```diff
@@ -174,3 +182,139 @@ func ProtobufLabelsFromAPILabels(labels map[string]string) []*cwssaws.Label {
...
+	var available uint64
+	if usableHosts > ifaceCount {
+		available = usableHosts - ifaceCount
+	}
+	return &cipam.Usage{
+		AvailableIPs:              available,
+		AcquiredIPs:               ifaceCount,
+		AvailableSmallestPrefixes: 0,
+		AvailablePrefixes:         nil,
+		AcquiredPrefixes:          instCount,
```
bcavnvidia: What is this telling the user?

bcavnvidia: The number of prefixes that are actively providing IPs to instances?

api/pkg/api/handler/vpcprefix.go
hwadekar-nv: `includeUsageStats` can be written common function but `queryIncludeRelations` can be used individual realationName purpose, we can move   `includeUsageStats` common

api/pkg/api/model/util/util.go
hwadekar-nv: Make sense to return `0` instead returning error,

db/pkg/db/ipam/ipam.go
hwadekar-nv: number of prefixes that haven't been totally exhausted, it seems correct way say

api/pkg/api/model/util/util.go
hwadekar-nv: I believe that’s accurate. The number of prefixes currently being acquired or consumed by instances is what we’re interested in. The idea is that we’re not using IPAM for all these prefixes now. We wanted to verify if this is the correct way to do it. Especially when we create interfaces at CORE for this instance, how many IPs do we use (for example, gateway/broadcast, etc.) for this interface so that we can calculate the total count based on that?

thossain-nv: @hwadekar-nv `AcquiredIPs` should be sum of all IPs used by all Interfaces. @bcavnvidia Each Interface consumes more than 1 IP, correct? And this differs for Subnets and VPC Prefix based interfaces.

Similarly `AcquiredPrefixes` should reflect the distinct prefixes used by all Instances under the Subnet/VPC Prefixes. If each Interface gets their own prefix, then this could be the total Interface count.

db/pkg/db/ipam/ipam.go
thossain-nv: `AvailablePrefixes` is typically an array, showing the largest prefix or prefixes available within the parent prefix. This would require inspecting each Interface prefixes and we should leverage the IPAM library to feed Interface prefixes into a Subnet/VPC Prefix prefix object and then retrieve the stats.

ajf:
@hwadekar-nv you have conflicts, can we get this merged?  Is there anything left?

cc @thossain-nv

hwadekar-nv:
Thanks @ajf , will rebase and after review, we will merge this week

api/pkg/api/model/util/util.go
hwadekar-nv: Fixed it in latest commit

thossain-nv (approved):
Looks good, thanks @hwadekar-nv

PR #484, opened 2026-05-05
fix: Bump pgx to v5.9.0 and mongo-driver to v1.17.7 for high-severity CVEs

kfelternv:
Resolves security vulnerability in pgx dependency by upgrading the package.


thossain-nv (approved):
Thanks @kfelternv

PR #490, opened 2026-05-06
fix: Clean up related components on Site deletion

nvlitagaki:
## Description
Ensure that when a site is deleted that we delete any records that may reference said site.

Also clean-up the database to deal with any records that were orphaned when sites were deleted previously.

## Type of Change

## Services Affected

## Breaking Changes

## Testing


nvlitagaki:
@coderabbitai review

workflow/pkg/activity/site/site.go
```diff
@@ -80,209 +83,478 @@ func (mst ManageSite) DeleteSiteComponentsFromDB(ctx context.Context, siteID uui
...
+	desdDAO := cdbm.NewDpuExtensionServiceDeploymentDAO(mst.dbSession)
+	tsDAO := cdbm.NewTenantSiteDAO(mst.dbSession)
+	skuDAO := cdbm.NewSkuDAO(mst.dbSession)
+	aDAO := cdbm.NewAllocationDAO(mst.dbSession)
+	acDAO := cdbm.NewAllocationConstraintDAO(mst.dbSession)
+	esDAO := cdbm.NewExpectedSwitchDAO(mst.dbSession)
+	epsDAO := cdbm.NewExpectedPowerShelfDAO(mst.dbSession)
+	emDAO := cdbm.NewExpectedMachineDAO(mst.dbSession)
+
+	// Run all of the cleanup work in a single transaction so that a failure
```
thossain-nv: Clean up should be best effort, I don't think we should roll back the clean up.

workflow/pkg/activity/site/site.go
```diff
@@ -80,209 +83,478 @@ func (mst ManageSite) DeleteSiteComponentsFromDB(ctx context.Context, siteID uui
...
+
+		for _, sku := range skus {
+			serr := skuDAO.Delete(ctx, tx, sku.ID)
+			if serr != nil && serr != cdb.ErrDoesNotExist {
+				logger.Error().Err(serr).Str("SKU ID", sku.ID).Msg("error deleting SKU record in DB")
+				return serr
+			}
+		}
+
+		// Delete Allocations
```
thossain-nv: Sites cannot be deleted while Allocations exist. So we don't need to re-delete them here.

workflow/pkg/activity/site/site.go
```diff
@@ -80,209 +83,478 @@ func (mst ManageSite) DeleteSiteComponentsFromDB(ctx context.Context, siteID uui
...
+			}
+		}
 
-	for _, ibp := range ibps {
-		// Delete IB Partition
-		serr := ibpDAO.Delete(ctx, nil, ibp.ID)
-		if serr != nil && serr != cdb.ErrDoesNotExist {
-			logger.Error().Err(serr).Str("IB Partition ID", ibp.ID.String()).Msg("error deleting IB Partition record in DB")
-			return serr
+		// Delete Tenant Sites
```
thossain-nv: Tenant Site deletion is tied to Allocation deletion, so this might also be redundant.

workflow/pkg/activity/site/site.go
```diff
@@ -80,209 +83,478 @@ func (mst ManageSite) DeleteSiteComponentsFromDB(ctx context.Context, siteID uui
...
 
-	// Delete Instances
-	// Check that Instance exists
-	instances, _, err := instanceDAO.GetAll(ctx, nil, cdbm.InstanceFilterInput{SiteIDs: []uuid.UUID{siteID}}, cdbp.PageInput{Limit: cdb.GetIntPtr(cdbp.TotalLimit)}, nil)
-	if err != nil {
-		logger.Error().Err(err).Msg("failed to retrieve Instances from DB by Site ID")
-		return err
-	}
+		// Delete InfiniBand interfaces
+		err = ibiDAO.DeleteAllBySiteID(ctx, tx, siteID)
```
thossain-nv: Clean up of Ethernet, InfiniBand, NVLink Interfaces & SSH Key Group Instance associations are already handled in the code that deletes the Instance (fixed recently https://github.com/NVIDIA/infra-controller-rest/blob/main/workflow/pkg/activity/instance/instance.go#L961) so this might also be redundant.

Any lingering un-deleted items will be handled by the migration added in this PR.

thossain-nv (commented):
Thanks for initiating this @nvlitagaki Added some suggestion to simplify the PR.

workflow/pkg/activity/site/site.go
nvlitagaki: Sounds good, thanks!

workflow/pkg/activity/site/site.go
nvlitagaki: Confirmed that deletion of the tenant site is handled atomically with deletion of the tenant's last allocation on that site.

workflow/pkg/activity/site/site.go
nvlitagaki: @thossain-nv Shouldn't we be worried about instances that were cleaned up earlier within thisDeleteSiteComponentsFromDb activity? Those wouldn't be affected by the recent fix... My assumption is that the instances in question would be invariably be those created via targeted instance creation, since any other instances would have had to have been deleted to clear the way for allocation deletion as a prerequisite for deleting the site

workflow/pkg/activity/site/site.go
```diff
@@ -80,13 +83,31 @@ func (mst ManageSite) DeleteSiteComponentsFromDB(ctx context.Context, siteID uui
...
+	skgiaDAO := cdbm.NewSSHKeyGroupInstanceAssociationDAO(mst.dbSession)
+	nsgDAO := cdbm.NewNetworkSecurityGroupDAO(mst.dbSession)
+	desdDAO := cdbm.NewDpuExtensionServiceDeploymentDAO(mst.dbSession)
+	tsDAO := cdbm.NewTenantSiteDAO(mst.dbSession)
+	skuDAO := cdbm.NewSkuDAO(mst.dbSession)
+	aDAO := cdbm.NewAllocationDAO(mst.dbSession)
+	acDAO := cdbm.NewAllocationConstraintDAO(mst.dbSession)
+	esDAO := cdbm.NewExpectedSwitchDAO(mst.dbSession)
+	epsDAO := cdbm.NewExpectedPowerShelfDAO(mst.dbSession)
+	emDAO := cdbm.NewExpectedMachineDAO(mst.dbSession)
```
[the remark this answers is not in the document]
nvlitagaki: Per Tareque, we are just going for a best effort clean-up and not worrying about rollback.

nvlitagaki:
@coderabbitai review

nvlitagaki:
@coderabbitai review

workflow/pkg/activity/site/site.go
```diff
@@ -237,6 +284,37 @@ func (mst ManageSite) DeleteSiteComponentsFromDB(ctx context.Context, siteID uui
 		}
 	}
 
+	// Delete VPC Prefixes
+	vpcPrefixes, _, err := vpfxDAO.GetAll(ctx, nil, cdbm.VpcPrefixFilterInput{SiteIDs: []uuid.UUID{siteID}}, cdbp.PageInput{Limit: cdb.GetIntPtr(cdbp.TotalLimit)}, nil)
```
hwadekar-nv, line 287: Thanks @nvlitagaki, Should we consider the order of resources for deletion here? (Higher importance should be prioritized at the end.)

workflow/pkg/activity/site/site.go
```diff
@@ -278,6 +356,188 @@ func (mst ManageSite) DeleteSiteComponentsFromDB(ctx context.Context, siteID uui
...
+			NetworkSecurityGroupID: nsg.ID,
+		})
+		if serr != nil && serr != cdb.ErrDoesNotExist {
+			logger.Error().Err(serr).Str("Network Security Group ID", nsg.ID).Msg("error deleting Network Security Group record in DB")
+			return serr
+		}
+	}
+
+	// Delete operating system site associations. After the associations for this site are removed, walk the set of
+	// operating systems they referenced and delete any OS whose only remaining site association was the one we just
```
thossain-nv: We want to keep the parent OS structure. We also don't delete SSH Key Groups when all Site Associations are removed. The user can decide whether to add some other Sites to it or remove it completely.

thossain-nv (approved):
Thank you for the changes @nvlitagaki, one last suggestion before merging.

Right now image based OSes are 1:1 so this change makes sense but it won't be the case with Patrice's work that I'm finishing up.

PR #501, opened 2026-05-07
chore: Move VPC update/delete proto building onto the DB model

chet:
## Description

Adds receiver methods on `cdbm.Vpc` so handlers don't have to build update/delete workflow protos inline:

- `(*Vpc).ToUpdateRequestProto()` and `ToDeletionRequestProto()` -- used by the Update and Delete handlers respectively (we do something similar for `Tenant`).
- `(*Vpc).GetSiteID()`, which replaces the `common.GetSiteVpcID` helper.

Tests cover ID-vs-ControllerVpcID fallback, nil description, nil labels, and nil NSG.

This is part of a wider effort I'm going to slowly be churning on for standardizing `ToProto` / `FromProto` across the board (which started with https://github.com/NVIDIA/infra-controller-rest/pull/500, where I'm also dropping an `AGENTS.md` which mentions something similar).

Signed-off-by: Chet Nichols III <chetn@nvidia.com>

## Type of Change

- [ ] **Feature** - New feature or functionality (feat:)
- [ ] **Fix** - Bug fixes (fix:)
- [x] **Chore** - Modification or removal of existing functionality (chore:)
- [ ] **Refactor** - Refactoring of existing functionality (refactor:)
- [ ] **Docs** - Changes in documentation or OpenAPI schema (docs:)
- [ ] **CI** - Changes in GitHub workflows. Requires additional scrutiny (ci:)
- [ ] **Version** - Issuing a new release version (version:)

## Services Affected

- [x] **API** - API models or endpoints updated
- [x] **Workflow** - Workflow service updated
- [x] **DB** - DB DAOs or migrations updated
- [ ] **Site Manager** - Site Manager updated
- [ ] **Cert Manager** - Cert Manager updated
- [ ] **Site Agent** - Site Agent updated
- [ ] **RLA** - RLA service updated
- [ ] **Powershelf Manager** - Powershelf Manager updated
- [ ] **NVSwitch Manager** - NVSwitch Manager updated

## Breaking Changes
- [ ] This PR contains breaking changes

## Testing

- [ ] Unit tests added/updated
- [ ] Integration tests added/updated
- [ ] Manual testing performed
- [ ] No testing required (docs, internal refactor, etc.)

PR #505, opened 2026-05-07
chore: Implementing layered proto conversion for VPC handlers

chet:
## Description

Implements a layered convention for proto-conversion methods that the rest of the proto handling rollout will adopt. This was already done for `Expected*` components, but VPC had some additional bits that I was fiddling with (I did #501 but that was some stuff to be shared across `vpc`, `vpcprefix`, and `vpcpeering`).

All said, this splits proto conversion into clear layers, and applies to `vpc` handlers:
- A primary `ToProto`/`FromProto` on the DB entity.
- Followed by `ToProto` on each API request type for request-shape conversion.
- With entity-level methods for handlers with no API request body (e.g. delete).

`AGENTS.md` has been further updated to document this layering.

Tests added!

## Type of Change

## Services Affected

## Breaking Changes

## Testing

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -183,32 +183,80 @@ make kind-down              # tear down cluster
 
 DB and API model types that round-trip with a workflow-schema (`cwssaws`)
 or Flow (`flowv1`) protobuf types carry conversion as receiver methods, not
-free functions. The naming and shape are uniform so call sites are
-predictable:
+free functions. The convention layers cleanly so call sites are
+predictable and every entity has the same surface:
 
-1. **One model type ↔ one proto type:** `func (m *T) ToProto(...) *protoT`
-   and `func (m *T) FromProto(p *protoT, ...)`. `FromProto` mutates the
-   receiver, treats a `nil` proto as a no-op, and returns no error —
-   callers pre-validate anything risky like UUID strings, and the method
-   leaves the receiver field unchanged on parse failure.
-2. **Side inputs that are not on the model** (BMC credentials, a linked
-   machine ID resolved by the caller, a fallback timestamp) are passed as
-   additional arguments — preferably grouped into a `XCredentials` struct
-   declared next to the model, with a comment explaining why the field
-   isn't persisted.
-3. **One model type → multiple proto request types:** when the same
-   record produces, for example, both a Create and an Update request, use
-   `ToCreateRequestProto()` / `ToUpdateRequestProto()` (see `Tenant`).
-4. **Sub-messages of a proto request:** when a request DTO produces a
+1. **Primary entity ↔ proto entity** lives on the DB model:
+   `func (m *T) ToProto(...) *protoT` and
+   `func (m *T) FromProto(p *protoT, ...)` — symmetric pair, defined
+   together. `FromProto` mutates the receiver and returns no error.
+   The field-level contract is:
+   - A `nil` proto is a no-op (receiver untouched).
+   - Required ID fields (e.g. `m.ID`) are preserved on a missing or
+     unparseable proto value, because callers pre-validate UUIDs
+     before calling.
+   - Optional pointer fields are cleared when the proto omits them
+     **or** when the proto value is invalid (e.g. an unparseable
+     UUID). For example, `(*Vpc).FromProto` clears
+     `NVLinkLogicalPartitionID` in both cases so the receiver ends
+     up as a clean reset rather than a partial merge.
+2. **Per-API-request → proto request** lives on the corresponding API
+   request type, not on the entity:
+   `func (req *APIXCreateRequest) ToProto(...) *protoXCreateRequest`,
+   `func (req *APIXUpdateRequest) ToProto(...) *protoXUpdateRequest`.
+   These methods commonly read the canonical fields via the entity's
+   `ToProto()` (passed in or fetched) and overlay request-specific
+   fields. Putting them on the API request type keeps the entity
+   surface focused on the canonical representation. When the API
+   request type uses `*T` to distinguish "field not provided" (nil)
+   from "explicitly clear" (non-nil pointer to zero value), the
+   request's `ToProto` is responsible for preserving that distinction
+   on the wire — overriding the entity-derived value when the API
+   request touched the field, even if the post-merge entity has the
+   same zero value.
+
+   **`ToProto` does not return errors.** It trusts that the request
+   has already been validated and that the handler has performed any
+   cross-context checks Validate cannot see. Validation lives in
+   `(req *APIXCreateRequest).Validate() error`, which is the
+   universal pre-`ToProto` step and must be called before the request
+   reaches `ToProto`. Anything `ToProto` would otherwise have to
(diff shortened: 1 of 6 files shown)


db/pkg/db/model/vpc.go
```diff
@@ -148,6 +148,71 @@ func (vpc *Vpc) toMetadataProto() *cwssaws.Metadata {
...
+	if proto.DefaultNvlinkLogicalPartitionId != nil {
+		if id, err := uuid.Parse(proto.DefaultNvlinkLogicalPartitionId.Value); err == nil {
+			vpc.NVLinkLogicalPartitionID = &id
+		} else {
+			vpc.NVLinkLogicalPartitionID = nil
+		}
+	} else {
+		vpc.NVLinkLogicalPartitionID = nil
+	}
+	if proto.Metadata != nil {
```
hwadekar-nv, line 224: Can we check name from Metadata in case if `proto.Name` is nill or empty?

api/pkg/api/handler/vpc.go
```diff
@@ -398,41 +397,14 @@ func (cvh CreateVPCHandler) Handle(c echo.Context) error {
...
-			RoutingProfileType:        routingProfile,
-			NetworkSecurityGroupId:    vpc.NetworkSecurityGroupID,
-			Vni:                       vni,
-		}
-
-		// Add default NVLinkLogicalPartition ID if it is present
-		if vpc.NVLinkLogicalPartitionID != nil {
-			createVpcRequest.DefaultNvlinkLogicalPartitionId = &cwssaws.NVLinkLogicalPartitionId{Value: vpc.NVLinkLogicalPartitionID.String()}
-		}
+		createVpcRequest := apiRequest.ToProto(vpc, &nwvt, vni, routingProfile)
```
thossain-nv: Should we move `vni` and `routingProfile` conversion into `ToProto`? They were done in the handler because we didn't have the `ToProto`.

api/pkg/api/model/vpc.go
```diff
@@ -208,6 +229,23 @@ func (asur APIVpcUpdateRequest) Validate() error {
...
+// further set `DefaultNvlinkLogicalPartitionId` on the returned
+// request when the API request changes that field.
+func (asur APIVpcUpdateRequest) ToProto(vpc *cdbm.Vpc) *cwssaws.VpcUpdateRequest {
+	vpcProto := vpc.ToProto()
+	return &cwssaws.VpcUpdateRequest{
+		Id:                     vpcProto.Id,
+		NetworkSecurityGroupId: vpcProto.NetworkSecurityGroupId,
+		Metadata:               vpcProto.Metadata,
+	}
+}
```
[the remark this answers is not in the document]
thossain-nv: Very esoteric catch @coderabbitai. @chet this seems like a valid concern?

db/pkg/db/model/vpc.go
thossain-nv: Right, we should fill the name from proto.Metadata since we're deprecating the top level proto name attribute.

thossain-nv (commented):
Beautiful pattern @chet, a few thoughts.

api/pkg/api/handler/vpc.go
chet: @thossain-nv Done for `vni ` and `nwvt`. Let me see how I can fiddle with `routingProfile` to make it work.

db/pkg/db/model/vpc.go
chet: Yup good call. Done!

api/pkg/api/handler/vpc.go
chet: @thossain-nv Okay, I think we're good in here now!

Implemented it as something akin to `TryFrom` in Rust, in that `ToProto` in this case returns a `(val, err)`, which will be used for cases where we need to do additional validation/conversion logic within the handler.

If we *want*, we call it something like `TryToProto` to make it obvious these are cases where we expect errors, or just have certain `ToProto` return err?

In this specific case, as part of moving `routingProfile` into `ToProto`, it returns `ErrRoutingProfileRequiresFNN` when a routing profile is set but the entity's network virtualization is not FNN.

But now we have:
```
createVpcRequest, derr := apiRequest.ToProto(vpc)
```

I considered making it `perr` but it seems like we want to keep `derr` so I left it.

Updated `AGENTS.md` to call this out as something we'll do as part of streamlining `ToProto/FromProto` also.

AGENTS.md
```diff
@@ -183,33 +183,65 @@ make kind-down              # tear down cluster
...
+   `ToProto()` (passed in or fetched) and overlay request-specific
+   fields. Putting them on the API request type keeps the entity
+   surface focused on the canonical representation. When the API
+   request type uses `*T` to distinguish "field not provided" (nil)
+   from "explicitly clear" (non-nil pointer to zero value), the
+   request's `ToProto` is responsible for preserving that distinction
+   on the wire — overriding the entity-derived value when the API
+   request touched the field, even if the post-merge entity has the
+   same zero value. May return an error when a mechanical conversion
+   on a request field can fail (e.g. integer width conversion).
```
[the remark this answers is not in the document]
chet: @coderabbitai Updated. I think we're good here? Resolve if so. Thanks!

api/pkg/api/model/vpc.go
```diff
@@ -208,6 +237,43 @@ func (asur APIVpcUpdateRequest) Validate() error {
...
 
+// ToProto builds the workflow request that pushes this Update's
+// merged-into-DB state to a Site. The persisted `vpc` is the source of
+// the wire fields because the handler has already merged the request's
+// (sparse) update fields into the entity by the time this is called;
+// sending the post-merge state matches the pre-existing handler
+// behaviour and keeps unchanged fields populated.
+//
+// `*string` and `*NVLinkLogicalPartitionId` overrides are applied for
+// `NetworkSecurityGroupID` and `NVLinkLogicalPartitionID` so the
```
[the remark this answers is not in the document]
chet: @coderabbitai Updated. I think we're good here? Resolve if so. Thanks!

thossain-nv (approved):
`ToProto` should trust the object, various validation happens before converting to proto, we want to consolidate that logic in `Validate` when possible.

api/pkg/api/handler/vpc.go
chet: Okay, talked to @thossain-nv.

Desired approach here is:

```
createVpcRequest := apiRequest.ToProto(vpc)
```

Errors shouldnt happen in `ToProto` (i.e. let's not have a `TryFrom` approach). Instead, lets leverage `Validate` across the board to validate before we attempt to `ToProto`. Validation will do any checking that would happen before `ToProto` (i.e. keep validation centralized), return an error if needed, and that means we are free/safe to call `ToProto` once we get to that point.

Adjusting now!

chet:
> `ToProto` should trust the object, various validation happens before converting to proto, we want to consolidate that logic in `Validate` when possible.

Done (as per comments in the thread above) -- thanks for the feedback/discussion to get this dialed in!

PR #517, opened 2026-05-11
feat: Improve TUI forms for instance create, instance update, and ssh-key-group create

kfelternv:
The TUI was missing some commands like instance update and instance reboot. Those are added now along with a few QOL improvements to the TUI pickers while running these commands.


thossain-nv (approved):
Looks good, thanks @kfelternv

PR #518, opened 2026-05-11
fix: Reject unknown --output values instead of silently using JSON

kfelternv:
Adds strict validator checks on --output to prevent silent failures falling back to json and creating unexpected script behavior


thossain-nv (approved):
Looks good, thanks @kfelternv

PR #521, opened 2026-05-11
fix: Avoid unsafe quoting in rackopreport JSON fallback

kfelternv:
Resolves unsafe quoting critical security finding https://github.com/NVIDIA/infra-controller-rest/security/code-scanning/43

### DIFF
--- a/flow/pkg/common/rackopreport/rackopreport.go
+++ b/flow/pkg/common/rackopreport/rackopreport.go
@@ -19,7 +19,6 @@ package rackopreport
 
 import (
 	"encoding/json"
-	"fmt"
 
 	"github.com/google/uuid"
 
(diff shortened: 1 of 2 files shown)


flow/pkg/common/rackopreport/rackopreport.go
```diff
@@ -137,16 +136,32 @@ func (rr *RackOpReport) Finalize() string {
...
+// fallbackErrorJSON returns a JSON-safe error report string used when
+// marshaling the FinalizedReport unexpectedly fails. The error message is
+// embedded via json.Marshal so any double quotes, backslashes, or control
+// characters in err are properly escaped and the returned value is always
+// valid JSON.
+func fallbackErrorJSON(err error) string {
+	payload := struct {
+		Error string `json:"error"`
+	}{
+		Error: "Failed to marshal report to JSON: " + err.Error(),
```
hwadekar-nv, line 156: Do we need to return an error here? Instead, I think we should return the original error.

hwadekar-nv (approved):

Issue #512, opened 2026-05-11
Improve TUI parity and surface machine blocking errors


kfelternv:
# Summary

The following is a list of problems I ran into while using the nico cli to investigate a nico deployment, that I asked AI to generate the issue description from all the context of the session. Everything below is AI Generated.

## Summary (AI)

The interactive TUI is missing several commands that exist in the generated non-interactive CLI, and machine error states are difficult to interpret because the actionable health alert is not visible in `machine list` and is buried inside full machine detail output.

We should bring the TUI closer to generated CLI parity and make machine-level blocking errors obvious directly in `machine list`, with more detail available in `machine get`.

## Problem

When investigating a machine in `Error`, the TUI currently shows only high-level fields like name, status, site, VPC, labels, and ID. The reason the machine is blocked can be hidden in `health.alerts`, for example:

- `BmcExplorationFailure`
- `classifications: ["PreventAllocations"]`
- target BMC endpoint
- message indicating the Redfish endpoint refused connections

That data is the key operational signal. Users should not have to leave the TUI, run non-interactive commands, pipe through `jq`, and manually inspect nested JSON just to understand why a machine cannot be allocated.

Separately, some generated CLI commands are unavailable in the TUI, including at least:

- `expected-rack`
- `expected-switch`
- `expected-power-shelf`
- `service-account get`

There are also TUI list commands that need owner-scope filters in service-account mode. For example, `allocation list` can fail unless `infrastructureProviderId` or `tenantId` is supplied.

## Requested changes

Add TUI support for missing generated CLI resources where practical:

- `expected-rack list/get`
- `expected-switch list/get`
- `expected-power-shelf list/get`
- `service-account get`

Improve `machine list` for machines in `Error`:

- Add a clear error/blocked column, for example `BLOCKED BY` or `HEALTH ALERT`.
- If `health.alerts[].classifications` contains `PreventAllocations`, show that alert's `id` in the list row.
- Include the alert target when it is concise enough to fit, for example `BmcExplorationFailure 10.91.x.x`.
- Prefer the most actionable blocking alert over generic status text.
- Keep healthy/ready machine rows concise and avoid noisy success probe data.

Improve `machine get` for machines in `Error`:

- Surface blocking health alerts near the top of the output.
- Show the alert `id`, target, classifications, and a shortened message before verbose JSON/detail output.
- Make it obvious whether `isUsableByTenant` is false because of a `PreventAllocations` health alert.

[...]

PR #534, opened 2026-05-15
feat: Add TUI parity commands and surface machine blocking alerts

kfelternv:
Adds various missing commands to the TUI that are already in the CLI. Also one quality of life improvement that will surface the blocking error on machines when listing them instead of making the user parse the json, parsing the json is still an option.


kfelternv:
Resolves #512

hwadekar-nv (approved):

PR #537, opened 2026-05-15
fix: Hint at api.name config when 404 source mismatches CLI config

kfelternv:
Based on a few bug reports, adding a small hint to direct users to the api.name configuration option would help debug 404s when using the cli with nico rest

### DIFF
--- a/cli/pkg/client.go
+++ b/cli/pkg/client.go
@@ -56,6 +56,12 @@ type APIError struct {
 	Body       string
 	Message    string
 	Data       interface{}
+	// Hint is an optional human-readable suggestion appended to the error
+	// message. Used to surface common-misconfiguration tips that the server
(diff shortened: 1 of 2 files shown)


thossain-nv (commented):
@kfelternv Could this confuse those who are using `/forge`?

thossain-nv (commented):
So the API will return a [json structure](https://github.com/NVIDIA/infra-controller-rest/blob/main/common/pkg/util/api.go#L51) for [404 responses](https://github.com/NVIDIA/infra-controller-rest/blob/main/api/pkg/middleware/notfound.go#L35) where the `source` field is set to the API name, we could just hint that?

thossain-nv (approved):
Looks good, thanks for the changes @kfelternv

thossain-nv:
@kfelternv Please migrate the PR to the unified repo.

PR #540, opened 2026-05-15
chore: Apply the ToProto modeling convention to InstanceType

chet:
## Description

Brings the `InstanceType` handler in line with our proto modeling. Instead of building workflow requests inline next to the RPC calls, the handler now goes through `ToProto`-style receiver methods on the appropriate types: the *create* and *update* flows live on the API request types (because each has a client body driving the shape), and the *delete* flow lives on the entity (because it has no API body).

**Capabilities** now ride along on the entity itself via a `Capabilities` relation, so a single `apiRequest.ToProto(it)` call is all the handler needs; per-capability wire rules live in `Validate`, keeping `ToProto` a pure mapper, which is the goal we're looking to achieve here.

We now have:
- API request side: `(*APIInstanceTypeCreateRequest).ToProto(it)` and `(*APIInstanceTypeUpdateRequest).ToProto(it)`
- Entity side: `(*InstanceType).ToProto()` / `FromProto(proto)`
- Capabilities are sourced from the new `it.Capabilities`, and each `(*MachineCapability).ToProto()` does its own mapping.

As mentioned above, per-capability rules (device type / InactiveDevices / numeric bounds) now live in `APIInstanceTypeCreateRequest.Validate` and `APIInstanceTypeUpdateRequest.Validate` (including a new `validateMachineCapUint32Bounds` that was put together), so by the time the request reaches `ToProto`, everything is safe to map.

I did have to do some reworking of things to fit the pattern, now it's all in line with the distinct `Validate` + `ToProto` flow that @thossain-nv is hoping to get out of this!

Tests added!

## Type of Change

- [x] **Refactor** - Refactoring of existing functionality (refactor:)

## Services Affected

- [x] **Site Agent** - Site Agent updated
- [ ] **Flow** - Flow service updated

## Breaking Changes

## Testing

### DIFF
--- a/api/pkg/api/handler/instancetype.go
+++ b/api/pkg/api/handler/instancetype.go
@@ -45,15 +45,12 @@ import (
 	cwu "github.com/NVIDIA/infra-controller-rest/workflow/pkg/util"
 
 	"github.com/NVIDIA/infra-controller-rest/api/internal/config"
-	"github.com/NVIDIA/infra-controller-rest/api/pkg/api/handler/util"
 	"github.com/NVIDIA/infra-controller-rest/api/pkg/api/handler/util/common"
 	"github.com/NVIDIA/infra-controller-rest/api/pkg/api/model"
-	mutil "github.com/NVIDIA/infra-controller-rest/api/pkg/api/model/util"
 	"github.com/NVIDIA/infra-controller-rest/api/pkg/api/pagination"
 	sc "github.com/NVIDIA/infra-controller-rest/api/pkg/client/site"
 	auth "github.com/NVIDIA/infra-controller-rest/auth/pkg/authorization"
 	cutil "github.com/NVIDIA/infra-controller-rest/common/pkg/util"
-	cwssaws "github.com/NVIDIA/infra-controller-rest/workflow-schema/schema/site-agent/workflows/v1"
 	"github.com/NVIDIA/infra-controller-rest/workflow/pkg/queue"
 )
 
@@ -246,7 +243,7 @@ func (cith CreateInstanceTypeHandler) Handle(c echo.Context) error {
 		}
 
 		// Get Machine capabilities for the Instance Type
-		mcs, _, derr = mcDAO.GetAll(ctx, tx, nil, []uuid.UUID{it.ID}, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, cdb.GetIntPtr(pagination.MaxPageSize), nil)
+		mcs, _, derr = mcDAO.GetAll(ctx, tx, nil, []uuid.UUID{it.ID}, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, cdb.GetIntPtr(cdbp.TotalLimit), nil)
 		if derr != nil {
 			logger.Error().Err(derr).Msg("error retrieving Machine capabilities for Instance Type")
 			return cutil.NewAPIError(http.StatusInternalServerError, "Failed to retrieve Machine capabilities for Instance Type", nil)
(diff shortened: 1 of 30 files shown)


db/pkg/db/model/machinecapability_test.go
```diff
@@ -1506,3 +1507,70 @@ func TestMachineCapability_GetIntInfo(t *testing.T) {
...
+		assert.Equal(t, "cpu-0", *proto.Name)
+		require.NotNil(t, proto.Count)
+		assert.Equal(t, uint32(4), *proto.Count)
+		require.NotNil(t, proto.Cores)
+		assert.Equal(t, uint32(16), *proto.Cores)
+		require.NotNil(t, proto.Threads)
+		assert.Equal(t, uint32(32), *proto.Threads)
+		assert.Nil(t, proto.DeviceType)
+		assert.Nil(t, proto.InactiveDevices)
+	})
```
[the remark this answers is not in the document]
chet: @coderabbitai I think we're good here? Lemme know. Thanks!

  Added three require.NotNil + assert.Equal pairs for proto.Frequency, proto.Vendor, proto.HardwareRevision in the CPU subtest of TestMachineCapability_ToProto. The test was already constructing the capability with those fields set, just wasn't checking they survived the conversion.

  Tests for the new behavior

  - instancetype_test.go (api model): two new Validate-table cases — negative InactiveDevices entry and MaxUint32+1 entry, both expecting wantErr: true.
  - instancetype_test.go (db model): one new TestInstanceType_FromProto subtest — "clears Name, Description, Labels when proto Metadata is nil" — sets stale values on the receiver, calls FromProto with Metadata: nil, asserts all three are cleared.

db/pkg/db/model/instancetype.go
```diff
@@ -95,6 +96,122 @@ type InstanceType struct {
...
+				labels[l.Key] = v
+			}
+			it.Labels = labels
+		} else {
+			it.Labels = nil
+		}
+	} else {
+		it.Description = nil
+		it.Labels = nil
+	}
```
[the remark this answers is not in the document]
chet: @coderabbitai Thanks! I think we're good here. Resolve if so, thanks!

api/pkg/api/model/instancetype.go
```diff
@@ -20,15 +20,104 @@ package model
...
+
+		if err := validateMachineCapUint32Bounds(mc.Count, "Count", mc.Name); err != nil {
+			return err
+		}
+		if err := validateMachineCapUint32Bounds(mc.Cores, "Cores", mc.Name); err != nil {
+			return err
+		}
+		if err := validateMachineCapUint32Bounds(mc.Threads, "Threads", mc.Name); err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
chet: @coderabbitai I think we're good here. Resolve if so, thanks!

Fwiw, inside the per-capability loop in validateMachineCapabilitiesForProto, added:
  for i, d := range mc.InactiveDevices {
      idx := d
      if err := validateMachineCapUint32Bounds(&idx, fmt.Sprintf("InactiveDevices[%d]", i), mc.Name); err != nil {
          return err
      }
  }

Soooo a request with InactiveDevices: [-1] now fails Validate with InactiveDevices[0] for capability "ib-0" must fit in uint32 (0..4294967295) instead of being silently cast to 4294967295 inside mc.ToProto().

api/pkg/api/handler/instancetype.go
```diff
@@ -259,148 +256,32 @@ func (cith CreateInstanceTypeHandler) Handle(c echo.Context) error {
...
-			}
+		it.Capabilities = make([]*cdbm.MachineCapability, len(mcs))
+		for i := range mcs {
+			it.Capabilities[i] = &mcs[i]
 		}
 
-		createInstanceTypeRequest.InstanceTypeAttributes = &cwssaws.InstanceTypeAttributes{
-			DesiredCapabilities: capabilities,
-		}
+		createInstanceTypeRequest := apiRequest.ToProto(it)
```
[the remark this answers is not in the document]
chet: @coderabbitai Is that better? Lemme know. Resolve if so. Thanks!

db/pkg/db/model/machinecapability_test.go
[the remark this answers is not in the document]
chet: @coderabbitai Oh and, resolve if it's good (and if you haven't resolved already at this point, haha). Thanks!

chet:
@coderabbitai Thanks!

api/pkg/api/model/instancetype.go
```diff
@@ -20,15 +20,110 @@ package model
...
+// and the entity / per-capability `ToProto` can be pure mappers.
+//
+// The rules:
+//   - Device type is only valid for Network (must be DPU) and GPU
+//     (must be NVLink); other capability types must not carry one.
+//   - InactiveDevices is only valid for InfiniBand capabilities.
+//   - Count / Cores / Threads must fit in uint32 so the proto numeric
+//     conversion is a safe cast.
+//   - Each (Type, Name) pair must be unique within the list.
+func validateMachineCapabilitiesForProto(caps []APIMachineCapability) error {
```
thossain-nv: So this validation should be taking place in `CreateInstanceTypRequest` and `UpdateInstanceTypeRequest`'s `Validate` method - which should in turn call `APIMachineCapability`'s `Validate` method in a loop or use `ozzo`'s validation helper for slices e.g. https://github.com/go-ozzo/ozzo-validation#mapsslicesarrays-of-validatables. This might produce a nested but accurate validation error structure.

api/pkg/api/model/instancetype.go
```diff
@@ -20,15 +20,110 @@ package model
...
+			}
+		}
+		for i, d := range mc.InactiveDevices {
+			idx := d
+			if err := validateMachineCapUint32Bounds(&idx, fmt.Sprintf("InactiveDevices[%d]", i), mc.Name); err != nil {
+				return err
+			}
+		}
+
+		if err := validateMachineCapUint32Bounds(mc.Count, "Count", mc.Name); err != nil {
```
thossain-nv: `ozzo` https://github.com/go-ozzo/ozzo-validation probably has some built in goodies to handle this so we don't have to write custom functions.

db/pkg/db/model/common.go
```diff
@@ -48,3 +48,21 @@ func LabelsFromProtoMetadata(md *cwssaws.Metadata) map[string]string {
 	}
 	return result
 }
+
+// LabelsToProto converts a string map of labels to the workflow proto
+// repeated Label representation. Returns nil for a nil map; an empty map
+// yields a non-nil empty slice so callers can distinguish "labels
+// explicitly cleared" from "no labels at all".
+func LabelsToProto(labels map[string]string) []*cwssaws.Label {
```
thossain-nv: It might be really nice to define a `struct APILabels map[string]string` and give it a `ToProto` receiver function!

db/pkg/db/model/instancetype.go
```diff
@@ -95,6 +96,109 @@ type InstanceType struct {
...
+	var capabilities []*cwssaws.InstanceTypeMachineCapabilityFilterAttributes
+	for _, mc := range it.Capabilities {
+		if mc == nil {
+			continue
+		}
+		capabilities = append(capabilities, mc.ToProto())
+	}
+	return &cwssaws.InstanceType{
+		Id:       it.ID.String(),
+		Metadata: it.toMetadataProto(),
```
thossain-nv: I would inline this instead of defining `toMetadataProto` Keeps the proto conversion logic contained within a single place.

db/pkg/db/model/machinecapability.go
```diff
@@ -78,6 +80,20 @@ var (
 	}
 )
 
+// intPtrToUint32Ptr converts a *int to *uint32. nil in, nil out.
+// Callers must ensure the int sits in `[0, MaxUint32]`; the cast
+// otherwise silently wraps. On the API-inbound path that's the job of
+// the request-side `Validate` (which rejects negatives and overflow
+// with a 400). On the proto-inbound path values originate from a
+// `uint32` field and are always non-negative.
+func intPtrToUint32Ptr(i *int) *uint32 {
```
thossain-nv: Might be nice to start a `common/pkg/util/converter.go` where we add these functions for all packages to use.

thossain-nv (commented):
Thank you for the changes @chet, added some suggestions to streamline the `Validate` and `ToProto`/`FromProto` functions.

api/pkg/api/model/instancetype.go
chet: @thossain-nv Done! The per-capability rules (`DeviceType` / `InactiveDevices` / numeric bounds) all moved onto `APIMachineCapability.Validate`, and both `APIInstanceTypeCreateRequest.Validate` and `APIInstanceTypeUpdateRequest.Validate` now apply them with `validation.Field(...)`.

The only thing left at the list level is `validateMachineCapabilityUniqueness`, which enforces the duplicate `(Type, Name)` check that can't be done per-element.

api/pkg/api/model/instancetype.go
chet: @thossain-nv Done! `validateMachineCapUint32Bounds` is gone. The `Count` / `Cores` / `Threads` bounds are now `validation.Min(0)` + `validation.Max(math.MaxUint32)` rules inside `APIMachineCapability.Validate`.

The `InactiveDevices` entry bounds were trickier, so they live in a `validation.By(...)` closure  — still inside `Validate`, though.

db/pkg/db/model/common.go
chet: @thossain-nv Ooooh hotness, yes! Done! Added `type Labels map[string]string` in `db/pkg/db/model/common.go` with `Labels.ToProto()`. I went with `Labels` rather than `APILabels` ...since the type lives in the DB-model package alongside the proto conversion, an API-prefixed name there *might* be misleading? But.. happy to rename if you'd prefer!

db/pkg/db/model/instancetype.go
chet: @thossain-nv  Done! `toMetadataProto` is gone, body is inlined into `ToProto`. With the `Labels.ToProto` introduction from above (boooyah), the metadata block in `ToProto` is now just a few lines (and keeps the whole proto conversion in one method).

db/pkg/db/model/machinecapability.go
chet: @thossain-nv Done! Added `common/pkg/util/converter.go` with `IntPtrToUint32Ptr` -- and `MachineCapability.ToProto` now calls it.

chet:
@thossain-nv I think I might do some more clean-up in here based on your feedback. I had done a fair a bit of twiddling to get it to where it was, and then subsequent twiddling based on your feedback, and now I'm thinking of a few more possible adjustments.

The `DeviceType` and `InactiveDevices` rules would read cleaner as named methods on `APIMachineCapability`, e.g.
- `validation.Field(&mc.DeviceType, validation.By(mc.validateDeviceType)),`
- `validation.Field(&mc.InactiveDevices, validation.By(mc.validateInactiveDevices)),`

It'd be the same logic, just hung off the type instead of as inline, and basically mirrors your general feedback with `APIMachineCapability.Validate` owning the rules.

Also, do something similar to `Labels.ToProto`, but for `APIMachineCapabilities []APIMachineCapability`, which gives us:
```
  func (caps APIMachineCapabilities) Validate() error {
      // delegates to each element's Validate, plus the uniqueness check
  }
  ```

..and then `APIInstanceTypeCreateRequest.MachineCapabilities` becomes `APIMachineCapabilities`, and `Validate` becomes:
```
  validation.Field(&itcr.MachineCapabilities),  // ..where ozzo calls APIMachineCapabilities.Validate()
```

And this is more of a simplification thing, but we could have `InstanceType.AttachCapabilities(mcs)` to handle some of the boilerplate stuff. Idea is right now, the create + update handlers both have:
```
  slices.SortFunc(mcs, func(a, b cdbm.MachineCapability) int { return a.Index - b.Index })
  it.Capabilities = make([]*cdbm.MachineCapability, len(mcs))
  for i := range mcs {
      it.Capabilities[i] = &mcs[i]
  }
  ```

..so it could become `it.AttachCapabilities(mcs)`, and the handler shrinks to one line.

Last one is tiny -- we'd just inline `util.ValidateLabels(itcr.Labels)` into struct validation via `validation.By`, and fold it the struct rules as `validation.Field(&itcr.Labels, validation.By(util.ValidateLabels))`, which keeps all field validation in one block.

chet:
@thossain-nv Made those tweaks as I mentioned above (in addition to addressing your feedback), and I also took your feedback into consideration and updated `AGENTS.md` as well to further establish guidelines around this.

db/pkg/db/model/common.go
```diff
@@ -27,15 +27,46 @@ var protoJsonUnmarshalOptions = protojson.UnmarshalOptions{
...
-func LabelsFromProtoMetadata(md *cwssaws.Metadata) map[string]string {
+// `Labels` map. Returns nil when md is nil or md.Labels is nil; returns
+// an empty map (not nil) when Labels is non-nil but empty, so callers
+// can distinguish "no labels reported" from "labels explicitly cleared".
+//
+// Returns the typed `Labels` for symmetry with `(Labels).ToProto()`.
+// Entities that haven't opted into the typed `Labels` field can still
+// assign the result directly — Go's assignability rule allows
+// converting between a named type and its identical underlying type.
+func LabelsFromProtoMetadata(md *cwssaws.Metadata) Labels {
```
thossain-nv: We can defer this to another PR but `func (l Labels) FromProto([]*cwssaws.Label)` might fit our patterns best. Then it gets called like:
```
labels := cdbm.Labels{}
labels.FromProto(protoLabels)
```

thossain-nv (approved):
These are beautiful patterns to build on, thank you so much @chet!

db/pkg/db/model/common.go
chet: @thossain-nv Oh yeah -- definitely can do that here. Love pushing the pattern all the way through. I figure we're only a couple of PRs in, so if we can improve the shape of the pattern with each PR, it'l allow the subsequent ones to flow even better. Thanks for the call out!

PR #549, opened 2026-05-19
chore: Add logic to retry Core/Flow gRPC connection

thossain-nv:
## Description

Currently Site Agent logic that initializes Core/Flow gRPC client only tries to connect to gRPC server once and then gives up. When this happens, Site Agent stateful set has to be restarted manually.

This PR adds support for continuous retries, allowing the client to recover if gRPC server comes back online.

## Type of Change

## Services Affected

- [x] **Site Agent** - Site Agent updated

## Related Issues (Optional)

None

## Breaking Changes

## Testing

- [x] Manual testing performed
- [x] No testing required (docs, internal refactor, etc.)

## Additional Notes

None


site-agent/pkg/components/managers/coregrpc/init.go
```diff
@@ -48,18 +50,24 @@ func (coregrpc *API) Init() {
...
 func (coregrpc *API) Start() {
-	ManagerAccess.Data.EB.Log.Info().Msg("Core gRPC: Starting the core gRPC client")
+	ManagerAccess.Data.EB.Log.Info().Msg("Core gRPC: Starting Core gRPC client manager")
 
-	// Create the client here
-	// Each workflow will check and reinitialize the client if needed
-	if err := coregrpc.CreateGrpcClient(); err != nil {
-		ManagerAccess.Data.EB.Log.Error().Msgf("Core gRPC: failed to create gRPC client: %v", err)
+	// Keep retrying to create the client until it succeeds
+	for {
```
kunzhao-nv, line 60: In `manager.go`, Managers.Start() calls this synchronously before Bootstrap, Orchestrator, and FlowGrpc. This loop blocks the whole agent until Core connects. Before this PR, a failed one-shot connect still allowed the Temporal worker to start. Want to confirm if this is an intended change.

Similar for flowgrpc/init.go L73.

site-workflow/pkg/grpc/client/flow_client.go
```diff
@@ -67,6 +67,12 @@ const (
 
 	// defaultCheckCertificateIntervalSeconds is the default interval to check for certificate changes
 	defaultCheckFlowCertificateIntervalSeconds = 15 * 60 // 15 minutes in seconds
+
+	// gRPC client default dial timeout
+	defaultFlowGrpcDialTimeoutSeconds = 5 // 5 seconds
+
+	// Flow gRPC client default retry interval
+	FlowGrpcConnectionRetryIntervalSeconds = 5 // 5 seconds
```
kunzhao-nv: Duplicate of FlowGrpcConnectionRetryIntervalSeconds in site-workflow/pkg/grpc/client/flow_client.go:75. May need to unify into a single source of truth.

site-agent/pkg/components/config/config_manager.go
```diff
@@ -73,21 +73,21 @@ func NewElektraConfig(utMode bool) *conftypes.Config {
...
 	}
-	if cSecOpt < int(client.InsecureGrpc) && cSecOpt > int(client.MutualTLS) {
-		cSecOpt = int(client.ServerTLS)
+	if coreGrpcSecOpt < int(client.InsecureGrpc) && coreGrpcSecOpt > int(client.MutualTLS) {
+		coreGrpcSecOpt = int(client.ServerTLS)
 	}
-	sOpt := 0
-	flag.IntVar(&sOpt, "coreGrpcSecureOptions", cSecOpt, "Core gRPC security option")
-	conf.CoreGrpc.Secure = client.SecureOptions(sOpt)
+	flag.IntVar(&coreGrpcSecOpt, "coreGrpcSecureOptions", coreGrpcSecOpt, "Core gRPC security option") // This is a no-op, just defines the flag
```
kunzhao-nv: conf.CoreGrpc.Secure is set before flag.Parse() (line ~316), so CLI -coreGrpcSecureOptions cannot override env. If only env is supported, assign after flag.Parse(); otherwise remove the "no-op" comment on line 89.

Similar in L141 of this file.

kunzhao-nv (approved):
nit: The PR description suggests recovery whenever Core/Flow come back online. This PR adds retry until the first successful connect in Start(), plus faster retry on cert-reload failure. It does not add explicit reconnect after a later RPC failure—UpdateGrpcClientState only updates health; transient outages may still recover via gRPC’s built-in reconnect, but there is no new application-level redial loop. Maybe  narrow the description to startup (and cert-reload) retry.

site-agent/pkg/components/managers/coregrpc/init.go
thossain-nv: This is a good point. We had Site Agent running and connecting to Temporal but not doing anything because it can't connect to Core. Do we want to allow normal execution in such case? If Site Agent can't connect to Core, we should not show it as healthy. What do you think?

site-workflow/pkg/grpc/client/flow_client.go
thossain-nv: Fixing.

site-agent/pkg/components/managers/coregrpc/init.go
```diff
@@ -48,18 +50,24 @@ func (coregrpc *API) Init() {
...
+// Start starts the Core gRPC client manager
 func (coregrpc *API) Start() {
-	ManagerAccess.Data.EB.Log.Info().Msg("Core gRPC: Starting the core gRPC client")
+	ManagerAccess.Data.EB.Log.Info().Msg("Core gRPC: Starting Core gRPC client manager")
 
-	// Create the client here
-	// Each workflow will check and reinitialize the client if needed
-	if err := coregrpc.CreateGrpcClient(); err != nil {
-		ManagerAccess.Data.EB.Log.Error().Msgf("Core gRPC: failed to create gRPC client: %v", err)
+	// Keep retrying to create the client until it succeeds
```
sydneyl-nvidia: Do we want to make this fail after some number of retries? Not saying we should, just asking the question, would never giving up hide a problem (I assume eventually logs will be consulted to diagnose connection failures reported against rest API, maybe it is ok the just keep trying but a crash loop pod status after 15 minutes might help someone deploying catch an issue before declaring things "ready" for rest API usage.

site-agent/pkg/components/managers/coregrpc/init.go
hwadekar-nv: Just a thought, We can also consider exponential retry or a configurable value. However, recently, we’ve encountered mostly issues when the secret path or certificate has been changed. Those problems with the loop can definitely be solved without restating the STS.

hwadekar-nv (approved):

site-agent/pkg/components/managers/coregrpc/init.go
thossain-nv: That's a good point @sydneyl-nvidia, I was just thinking that we should Error the pod after a certain number of retries. One issue is that we've seen Core being offline for some reason for 15 minutes and Site Agent wasn't able to catch up when it came back online.

What could be a good number and period of retry?

thossain-nv: Added support for backoff retry with timeout.

site-agent/pkg/components/config/config_manager.go
thossain-nv: Removed the no-op comment.

site-agent/pkg/components/managers/coregrpc/init.go
thossain-nv: Added exponential backoff retry with timeout.

PR #574, opened 2026-05-23
feat: Add Flat VPC type and instance `auto` interface support

### LINKED ISSUE #351: feat: Support no-dpu instances (instances that are not part of any VPC)
### Is this a new feature, an enhancement, or a change to existing functionality?

New Feature

### How would you describe the priority of this feature request

Medium

### Please provide a clear description of problem this feature solves

[nico-core](https://github.com/NVIDIA/ncx-infra-controller-core) already supports bare-metal instances that do not make use of DPUs.

They get ingested as usual, but without any DPU installation steps or configurations.
The actual NIC (or DPU in NIC mode) on the host communicates directly via underlay network, and gets an IP address on one of the underlay network segments.
That means also all software on the host would communicate to nico-core using underlay networking - both system owned software (scout) as well as the tenant OS.

For tenants using these hosts this means the following:
- They should not be able to specify any network configuration for any interface - because there is no configurable NIC.
- The hosts should not be part of any VPC

To support this model, we need to consider at least the following:
- The create instance request should not require a VPC ID and no interfaces (network segments or vpc prefixes)
- The `vpcId` field in the instance would become optional
- The `vpcId` query parameter for "get all instances" would not be able to explicitly search for instances that are not part of a VPC - because the absence is equivalent to "search for all instances". Maybe we need some other mechanism here.

We might also want to consider whether to just encode the VPC ID in general on a per-network-interface level instead of per instance - given nico-core supports different interfaces to be attached to different VPCs.

### Feature Description

see above

### Describe your ideal solution

_No response_

### Describe any alternatives you have considered

_No response_

### Additional context

_No response_

### Code of Conduct

- [x] I agree to follow NCX Infra Controller's Code of Conduct
- [x] I have searched the [open feature requests](https://github.com/NVIDIA/ncx-infra-controller-rest/issues?q=is%3Aopen+is%3Aissue+type:Enhancement,Epic) and have found no duplicates for this feature request

chet:
## Description

This closes https://github.com/NVIDIA/infra-controller-rest/issues/351, and is the REST-side counterpart to the Flat VPC and `auto` interface work that landed in Core (https://github.com/NVIDIA/infra-controller-core/pull/1775, https://github.com/NVIDIA/infra-controller-core/pull/1576, and https://github.com/NVIDIA/infra-controller-core/pull/1674), which together let tenants put instances on zero-DPU hosts (or hosts with their DPU in NIC mode).

The discussion on #351 was originally about whether to model this as `vpcId: null` or a special "unmanaged VPC" type. What landed in Core is closer to the latter -- Flat VPCs are real VPCs with VNIs, NSGs, and peering, but NICo doesn't drive their data plane.

Changes included:
- Brought the two new fields into `nico_nico.proto`.
- `Flat` is a new value in the `network_virtualization_type` enum across the API, DB, OpenAPI, and proto layers.
- `auto` is a new boolean on instance create / update that flips the interface configuration model from "explicit list" to "NICo resolves from HostInband segments." Mutually exclusive with `interfaces`, persisted on the instance row so partial updates can re-issue the signal without the caller re-supplying it.
- Defense-in-depth check at the REST layer that `auto: true` requires a Flat VPC. Core enforces the same rule, but we fail-fast here too to avoid round-tripping the site for an obviously bad request.

This also adds some "VPC capability-like" mappings (similar to what we have in Core) to drive logic decisions at the REST API layer.

Tests added!

## Type of Change

## Services Affected

- [ ] **Flow** - Flow service updated

## Breaking Changes

## Testing

### DIFF
--- a/api/pkg/api/handler/instance.go
+++ b/api/pkg/api/handler/instance.go
@@ -70,6 +70,22 @@ type CreateInstanceHandler struct {
 	tracerSpan *cutil.TracerSpan
 }
 
+// buildInstanceNetworkConfig assembles the workflow
+// InstanceNetworkConfig from the persisted auto flag and the
+// per-interface configs built earlier in the handler. When auto is
+// true the explicit interface list is intentionally omitted: NICo
+// resolves interfaces from the host's HostInband segments, so
+// sending an explicit list alongside auto=true is contradictory
+// (rejected by Core, and on update could otherwise carry forward
+// the instance's previously-persisted interfaces).
+func buildInstanceNetworkConfig(auto bool, interfaceConfigs []*cwssaws.InstanceInterfaceConfig) *cwssaws.InstanceNetworkConfig {
+	nc := &cwssaws.InstanceNetworkConfig{Auto: auto}
+	if !auto {
+		nc.Interfaces = interfaceConfigs
+	}
+	return nc
+}
+
 // NewCreateInstanceHandler initializes and returns a new handler for creating Instance
 func NewCreateInstanceHandler(dbSession *cdb.Session, tc temporalClient.Client, scp *sc.ClientPool, cfg *config.Config) CreateInstanceHandler {
 	return CreateInstanceHandler{
@@ -363,6 +379,16 @@ func (cih CreateInstanceHandler) Handle(c echo.Context) error {
 	if vpc.ControllerVpcID == nil || vpc.Status != cdbm.VpcStatusReady {
 		logger.Warn().Msg("VPC specified in request data is not ready")
 		return cutil.NewAPIErrorResponse(c, http.StatusBadRequest, "VPC specified in request data is not ready", nil)
+	}
+
+	// `auto` is the explicit signal that an Instance lives on a zero-DPU
+	// host (or a host with its DPU in NIC mode); those instances must be
+	// in a Flat VPC. Core also rejects this combination, but we surface
+	// the error here as defense in depth and to avoid round-tripping the
+	// site for an obviously bad request.
+	if apiRequest.Auto && !cdbm.VpcTypeSupportsAutoInterface(vpc.NetworkVirtualizationType) {
+		logger.Warn().Msg("auto-network instances may only be created in a Flat VPC")
+		return cutil.NewAPIErrorResponse(c, http.StatusBadRequest, "`auto` is only supported when the VPC has `networkVirtualizationType` set to `FLAT`", nil)
 	}
 
 	var defaultNvllpID *uuid.UUID
@@ -1262,6 +1288,7 @@ func (cih CreateInstanceHandler) Handle(c echo.Context) error {
 			AlwaysBootWithCustomIpxe: *apiRequest.AlwaysBootWithCustomIpxe,
 			PhoneHomeEnabled:         *apiRequest.PhoneHomeEnabled,
 			UserData:                 apiRequest.UserData,
+			NetworkAuto:              apiRequest.Auto,
 			NetworkSecurityGroupID:   apiRequest.NetworkSecurityGroupID,
 			Labels:                   apiRequest.Labels,
 			IsUpdatePending:          false,
@@ -1286,7 +1313,7 @@ func (cih CreateInstanceHandler) Handle(c echo.Context) error {
 		// Update the controller ID
 		// We need this to match the instance ID.  This was previously handled
 		// by the async cloud workflow after successful creation on site.
-		instance, derr = instanceDAO.Update(ctx, tx, cdbm.InstanceUpdateInput{InstanceID: instance.ID, ControllerInstanceID: cdb.GetUUIDPtr(instance.ID)})
+		instance, derr = instanceDAO.Update(ctx, tx, cdbm.InstanceUpdateInput{InstanceID: instance.ID, InstanceUpdateCommon: cdbm.InstanceUpdateCommon{ControllerInstanceID: cdb.GetUUIDPtr(instance.ID)}})
 		if derr != nil {
 			logger.Error().Err(derr).Msg("unable to update Instance record controllerInstanceID in DB")
 			return cutil.NewAPIError(http.StatusInternalServerError, "Failed updating new Instance record, DB error", nil)
@@ -1552,10 +1579,8 @@ func (cih CreateInstanceHandler) Handle(c echo.Context) error {
 					TenantOrganizationId: tenant.Org,
 					TenantKeysetIds:      instanceSshKeyGroupIds,
 				},
-				Os: osConfig,
-				Network: &cwssaws.InstanceNetworkConfig{
-					Interfaces: interfaceConfigs,
-				},
+				Os:      osConfig,
+				Network: buildInstanceNetworkConfig(instance.NetworkAuto, interfaceConfigs),
 				Infiniband: &cwssaws.InstanceInfinibandConfig{
 					IbInterfaces: ibInterfaceConfigs,
 				},
@@ -1705,10 +1730,12 @@ func (uih UpdateInstanceHandler) handleReboot(c echo.Context, logger *zerolog.Lo
 		var derr error
 		ui, derr = instanceDAO.Update(ctx, tx,
 			cdbm.InstanceUpdateInput{
-				InstanceID:  instance.ID,
-				Name:        apiRequest.Name,
-				Description: apiRequest.Description,
-				PowerStatus: powerStatus,
+				InstanceID: instance.ID,
+				InstanceUpdateCommon: cdbm.InstanceUpdateCommon{
+					Name:        apiRequest.Name,
+					Description: apiRequest.Description,
+					PowerStatus: powerStatus,
+				},
 			},
 		)
 		if derr != nil {
@@ -2140,6 +2167,34 @@ func (uih UpdateInstanceHandler) Handle(c echo.Context) error {
 	// we know the caller has access to this instance.
 	if instance.Status == cdbm.InstanceStatusTerminating || instance.Status == cdbm.InstanceStatusTerminated {
 		return cutil.NewAPIErrorResponse(c, http.StatusConflict, "Instance is terminating and cannot be updated", nil)
+	}
+
+	// `auto: true` (re-)resolves interfaces from the host's HostInband
+	// segments; this only makes sense in a Flat VPC. Core also rejects
+	// this combination, but we surface the error here as defense in
+	// depth and to avoid round-tripping the site for an obviously bad
+	// request. `auto: false` is permitted on any VPC type (it's the
+	// default state).
+	if apiRequest.Auto != nil && *apiRequest.Auto && !cdbm.VpcTypeSupportsAutoInterface(vpc.NetworkVirtualizationType) {
+		logger.Warn().Msg("auto-network update is only valid for instances in a Flat VPC")
+		return cutil.NewAPIErrorResponse(c, http.StatusBadRequest, "`auto: true` is only supported when the Instance's VPC has `networkVirtualizationType` set to `FLAT`", nil)
+	}
+
+	// Reject explicit `interfaces` when the effective post-update state
+	// would still have auto enabled. `effectiveAuto` is the request's
+	// value when supplied, otherwise the instance's currently-persisted
+	// state. Validation already rejects (auto=true, interfaces=[...]) in
+	// the same payload, but a PATCH that omits `auto` against an
+	// already-auto instance would otherwise slip past and persist
+	// Interface rows that `buildInstanceNetworkConfig` then drops from
+	// the workflow payload -- leaving DB/site state diverged.
+	effectiveAuto := instance.NetworkAuto
+	if apiRequest.Auto != nil {
+		effectiveAuto = *apiRequest.Auto
+	}
+	if effectiveAuto && len(apiRequest.Interfaces) > 0 {
+		logger.Warn().Msg("explicit interfaces cannot be updated while the instance remains in auto mode")
+		return cutil.NewAPIErrorResponse(c, http.StatusBadRequest, "`interfaces` cannot be set while `auto` is true; disable `auto` first or omit `interfaces`", nil)
 	}
 
 	if instance.IsMissingOnSite {
@@ -2757,17 +2812,20 @@ func (uih UpdateInstanceHandler) Handle(c echo.Context) error {
 		var derr error
 		ui, derr = instanceDAO.Update(ctx, tx,
 			cdbm.InstanceUpdateInput{
-				InstanceID:               instanceID,
-				Name:                     apiRequest.Name,
-				Description:              apiRequest.Description,
-				OperatingSystemID:        osID,
-				IpxeScript:               apiRequest.IpxeScript,
-				AlwaysBootWithCustomIpxe: apiRequest.AlwaysBootWithCustomIpxe,
-				NetworkSecurityGroupID:   nsgID,
-				PhoneHomeEnabled:         apiRequest.PhoneHomeEnabled,
-				Status:                   instanceStatusConfiguring,
-				UserData:                 apiRequest.UserData,
-				Labels:                   apiRequest.Labels,
+				InstanceID: instanceID,
+				InstanceUpdateCommon: cdbm.InstanceUpdateCommon{
+					Name:                     apiRequest.Name,
+					Description:              apiRequest.Description,
+					OperatingSystemID:        osID,
+					IpxeScript:               apiRequest.IpxeScript,
+					AlwaysBootWithCustomIpxe: apiRequest.AlwaysBootWithCustomIpxe,
+					NetworkSecurityGroupID:   nsgID,
+					PhoneHomeEnabled:         apiRequest.PhoneHomeEnabled,
+					Status:                   instanceStatusConfiguring,
+					UserData:                 apiRequest.UserData,
+					NetworkAuto:              apiRequest.Auto,
+					Labels:                   apiRequest.Labels,
+				},
 			},
 		)
 		if derr != nil {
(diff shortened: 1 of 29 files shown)


api/pkg/api/handler/instance.go
```diff
@@ -1554,6 +1565,7 @@ func (cih CreateInstanceHandler) Handle(c echo.Context) error {
 				Os: osConfig,
 				Network: &cwssaws.InstanceNetworkConfig{
 					Interfaces: interfaceConfigs,
+					Auto:       apiRequest.Auto,
```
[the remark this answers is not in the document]
chet: @coderaitai Addressed. PTAL. Resolve if it looks good. Thanks!

db/pkg/db/model/instance.go
```diff
@@ -1021,6 +1030,13 @@ func (isd InstanceSQLDAO) UpdateMultiple(ctx context.Context, tx *db.Tx, inputs
 				isd.tracerSpan.SetAttribute(instanceDAOSpan, prefix+"user_data", *input.UserData)
 			}
 		}
+		if input.NetworkAuto != nil {
+			i.NetworkAuto = *input.NetworkAuto
+			columns = append(columns, "network_auto")
+			if addTrace {
+				isd.tracerSpan.SetAttribute(instanceDAOSpan, prefix+"network_auto", fmt.Sprintf("%t", *input.NetworkAuto))
+			}
+		}
```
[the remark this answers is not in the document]
chet: @coderaitai Addressed. PTAL. Resolve if it looks good. Thanks!

openapi/spec.yaml
```diff
@@ -16051,9 +16056,12 @@ components:
...
+          description: 'At least one interface must be specified unless `auto` is true. Either Subnet or VPC Prefix interfaces allowed. Only one of the Subnets or VPC Prefixes can be attached over Physical interface. If only one Subnet is specified, then it will be attached over physical interface regardless of the value of isPhysical. In case of VPC Prefix, isPhysical will always be true. Mutually exclusive with `auto`: when `auto` is true this list MUST be empty.'
           items:
             $ref: '#/components/schemas/InterfaceCreateRequest'
+        auto:
+          type: boolean
+          description: 'When true, asks NICo to auto-resolve the Instance''s network interfaces from the host''s underlay (HostInband) network segments. Intended for instances on zero-DPU hosts (or hosts with their DPU in NIC mode). When true, `interfaces` MUST be empty; the resolved interfaces surface on the Instance''s read response.'
```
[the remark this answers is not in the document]
chet: @coderaitai Addressed. PTAL. Resolve if it looks good. Thanks!

api/pkg/api/handler/instance.go
```diff
@@ -1554,6 +1565,11 @@ func (cih CreateInstanceHandler) Handle(c echo.Context) error {
 				Os: osConfig,
 				Network: &cwssaws.InstanceNetworkConfig{
 					Interfaces: interfaceConfigs,
+					// Mirror the persisted Instance state -- matches the
+					// update handler, which also sources `auto` from the
+					// DB record. The DAO may transform or default the
+					// request's value, so this is the safer source.
+					Auto: instance.NetworkAuto,
 				},
```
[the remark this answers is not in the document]
chet: @coderabbitai PTAL -- resolve if we're good. Thanks!

api/pkg/api/handler/instancebatch.go
```diff
@@ -419,6 +419,16 @@ func (bcih BatchCreateInstanceHandler) Handle(c echo.Context) error {
 		return cutil.NewAPIErrorResponse(c, http.StatusBadRequest, "VPC specified in request data is not ready", nil)
 	}
 
+	// `auto` is the explicit signal that an Instance lives on a zero-DPU
+	// host (or a host with its DPU in NIC mode); those instances must be
+	// in a Flat VPC. Core also rejects this combination, but we surface
+	// the error here as defense in depth and to avoid round-tripping the
+	// site for an obviously bad request.
+	if apiRequest.Auto && !cdbm.VpcTypeSupportsAutoInterface(vpc.NetworkVirtualizationType) {
```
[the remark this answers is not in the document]
chet: @coderabbitai PTAL -- resolve if we're good. Thanks!

api/pkg/api/handler/instance.go
```diff
@@ -2763,6 +2799,7 @@ func (uih UpdateInstanceHandler) Handle(c echo.Context) error {
 				PhoneHomeEnabled:         apiRequest.PhoneHomeEnabled,
 				Status:                   instanceStatusConfiguring,
 				UserData:                 apiRequest.UserData,
+				NetworkAuto:              apiRequest.Auto,
```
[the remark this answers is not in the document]
chet: @coderabbitai Feedback addressed. If it looks good, please resolve. Thanks!

db/pkg/db/model/instance.go
```diff
@@ -1078,35 +1110,39 @@ func (isd InstanceSQLDAO) UpdateMultiple(ctx context.Context, tx *db.Tx, inputs
...
+			partitions[signature] = &partition{
+				instances: make([]*Instance, 0),
+				columns:   append([]string{}, columns...),
+			}
 		}
+		partitions[signature].instances = append(partitions[signature].instances, i)
 
-	}
+		instances = append(instances, i)
+		ids = append(ids, input.InstanceID)
```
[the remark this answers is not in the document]
chet: @coderabbitai Feedback addressed. If it looks good, please resolve. Thanks!

sdk/standard/model_batch_instance_create_request.go
```diff
@@ -85,14 +87,13 @@ type _BatchInstanceCreateRequest BatchInstanceCreateRequest
...
 	this := BatchInstanceCreateRequest{}
 	this.NamePrefix = namePrefix
 	this.Count = count
 	this.TenantId = tenantId
 	this.InstanceTypeId = instanceTypeId
 	this.VpcId = vpcId
-	this.Interfaces = interfaces
 	var topologyOptimized bool = true
 	this.TopologyOptimized = &topologyOptimized
 	return &this
```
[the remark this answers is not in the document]
chet: @coderabbitai Feedback addressed. If it looks good, please resolve. Thanks!

sdk/standard/model_instance_create_request.go
```diff
@@ -84,12 +86,11 @@ type _InstanceCreateRequest InstanceCreateRequest
...
 // will change when the set of required properties is changed
-func NewInstanceCreateRequest(name string, tenantId string, vpcId string, interfaces []InterfaceCreateRequest) *InstanceCreateRequest {
+func NewInstanceCreateRequest(name string, tenantId string, vpcId string) *InstanceCreateRequest {
 	this := InstanceCreateRequest{}
 	this.Name = name
 	this.TenantId = tenantId
 	this.VpcId = vpcId
-	this.Interfaces = interfaces
 	return &this
 }
```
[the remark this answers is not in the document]
chet: @coderabbitai Feedback addressed. If it looks good, please resolve. Thanks!

api/pkg/api/handler/instance.go
```diff
@@ -2141,6 +2166,17 @@ func (uih UpdateInstanceHandler) Handle(c echo.Context) error {
...
+	// `auto: true` (re-)resolves interfaces from the host's HostInband
+	// segments; this only makes sense in a Flat VPC. Core also rejects
+	// this combination, but we surface the error here as defense in
+	// depth and to avoid round-tripping the site for an obviously bad
+	// request. `auto: false` is permitted on any VPC type (it's the
+	// default state).
+	if apiRequest.Auto != nil && *apiRequest.Auto && !cdbm.VpcTypeSupportsAutoInterface(vpc.NetworkVirtualizationType) {
+		logger.Warn().Msg("auto-network update is only valid for instances in a Flat VPC")
+		return cutil.NewAPIErrorResponse(c, http.StatusBadRequest, "`auto: true` is only supported when the Instance's VPC has `networkVirtualizationType` set to `FLAT`", nil)
+	}
```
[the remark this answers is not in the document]
chet: @coderabbitai Feedback addressed. If it looks good, please resolve. Thanks!

openapi/spec.yaml
```diff
@@ -15789,6 +15791,9 @@ components:
             - string
             - 'null'
           description: 'Serial Console URL for the Instance. Format: ssh://<id>@siteSerialConsoleHostname'
+        auto:
+          type: boolean
+          description: 'True when this Instance had its network interfaces auto-resolved by NICo from the host''s underlay (HostInband) network segments. When true, the caller''s request `interfaces` list was empty and `interfaces` here reflects the resolved set.'
```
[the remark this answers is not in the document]
chet: @coderabbitai Feedback addressed. If it looks good, please resolve. Thanks!

db/pkg/db/model/instance.go
```diff
@@ -885,10 +896,30 @@ func (isd InstanceSQLDAO) UpdateMultiple(ctx context.Context, tx *db.Tx, inputs
 		return []Instance{}, nil
 	}
 
-	// Build instances and collect columns to update
+	// Build instances and collect columns to update.
+	//
+	// Each Bulk UPDATE applies the same column list to every row it
+	// targets, so rows whose input did not specify a given column
+	// would silently be overwritten with the struct's zero value
+	// (e.g. `bool` -> false for nullable bool fields like
```
bcavnvidia: @coderabbitai This comment seems misaligned with the code.  UpdateMultiple accepts InstanceUpdateInput.  Aside from the UUID of the instance, all other fields of that struct are pointers, which have null as their zero-value, and null fields should represent "unset" and not be updated in this function.  Verify this, and then evaluate if the complexity being added by this partitioning logic still makes sense.

bcavnvidia: Ok, the inclusion of column X for one set of InstanceUpdateInput values and the absence of column X in a second InstanceUpdateInput means the second InstanceUpdateInput will see its column X set to the zero-value because the first InstanceUpdateInput is triggering it.

This feels like the approach for `Update` vs `UpdateMultiple` functions would benefit from a review and possible refactor.

In any case, the partitioning logic being here feels a little out of place, but the need is clear.

bcavnvidia: It might be better to separate the instance ID field out of the actual update fields and simply create separate structs for `InstanceUpdateInput` with a UUID  and `InstanceUpdateMultipleInput` with a slice of UUID, and then have both just embed a single `InstanceUpdateCommon` with the common fields.  The single-instance `Update` could then call  `UpdateMultiple` and set the UUID slice based on its single instance UUID.  Alternatively, to keep it simpler, this could be done by giving `InstanceUpdateMultipleInput` the UUID slice directly, and then embedding `InstanceUpdateMultipleInput` in `InstanceUpdateInput` and shadowing the UUID slice field in `InstanceUpdateInput` with a field that holds a single UUID.

bcavnvidia: For shadowing in approach 2, the expectation is that the field name would remain the same:

```
type InstanceUpdateInput struct {
    InstanceID uuid.UUID
    InstanceUpdateMultipleInput
}

// All IDs in InstanceIDs receive the same InstanceUpdateCommon values.
type InstanceUpdateMultipleInput struct {
    InstanceID []uuid.UUID
    ...
}
```

The downside would be the field name being singular in InstanceUpdateMultipleInput.

bcavnvidia: @coderabbitai Audit callers to check whether any existing call site relies on heterogeneous per-item fields.

bcavnvidia (approved):

db/pkg/db/model/instance.go
chet: @bcavnvidia answered your audit question + did the refactor you sketched:

- **Audit result:** no production caller of `UpdateMultiple` used heterogeneous per-row update masks. The only mixed-mask shape was a test (`batch_update_three_instances`) asserting the buggy behavior. The single existing batch caller (`instancebatch.go` setting `ControllerInstanceID = own ID` after CreateMultiple) uses heterogeneous *values* but a uniform *mask*.
- **Refactor:** split out `InstanceUpdateCommon` (the mask), kept `InstanceUpdateInput = { InstanceID; InstanceUpdateCommon }` for single-row, added `InstanceUpdateMultipleInput = { InstanceIDs []uuid.UUID; InstanceUpdateCommon }` for the shared-mask multi shape. Single-row `Update` is now a thin wrapper. The partition map / sort+strings imports / column-signature compare are gone — `UpdateMultiple` computes the column list once from the shared mask and does one Bulk per call.
- **Heterogeneous-values caller:** the post-CreateMultiple `ControllerInstanceID = inst.ID` loop in `instancebatch.go` now issues N single-row `Update` calls inside the existing transaction. N is capped at MaxBatchSize (~18), atomic, and the pattern matches what the single-instance create handler does.
- All struct-literal sites updated to the embedded shape.

Build, vet, and the touched test packages (`db/pkg/db/model`, `api/pkg/api/handler`, `api/pkg/api/model`, `workflow/pkg/activity/instance`) are green.

api/pkg/api/handler/instance.go
```diff
@@ -365,6 +381,16 @@ func (cih CreateInstanceHandler) Handle(c echo.Context) error {
 		return cutil.NewAPIErrorResponse(c, http.StatusBadRequest, "VPC specified in request data is not ready", nil)
 	}
 
+	// `auto` is the explicit signal that an Instance lives on a zero-DPU
+	// host (or a host with its DPU in NIC mode); those instances must be
+	// in a Flat VPC. Core also rejects this combination, but we surface
+	// the error here as defense in depth and to avoid round-tripping the
+	// site for an obviously bad request.
+	if apiRequest.Auto && !cdbm.VpcTypeSupportsAutoInterface(vpc.NetworkVirtualizationType) {
```
thossain-nv, line 389: `VpcTypeSupportsAutoInterface` seems to be doing validation and is better suited for `APIVpcCreateRequest`'s `ValidateRoutingProfile` receiver function.

api/pkg/api/handler/instance.go
```diff
@@ -1286,7 +1313,7 @@ func (cih CreateInstanceHandler) Handle(c echo.Context) error {
 		// Update the controller ID
 		// We need this to match the instance ID.  This was previously handled
 		// by the async cloud workflow after successful creation on site.
-		instance, derr = instanceDAO.Update(ctx, tx, cdbm.InstanceUpdateInput{InstanceID: instance.ID, ControllerInstanceID: cdb.GetUUIDPtr(instance.ID)})
+		instance, derr = instanceDAO.Update(ctx, tx, cdbm.InstanceUpdateInput{InstanceID: instance.ID, InstanceUpdateCommon: cdbm.InstanceUpdateCommon{ControllerInstanceID: cdb.GetUUIDPtr(instance.ID)}})
```
thossain-nv, line 1316: This should be called `InstanceUpdateCommonInput`

api/pkg/api/handler/instance.go
```diff
@@ -2942,8 +3000,30 @@ func (uih UpdateInstanceHandler) Handle(c echo.Context) error {
...
+		//     mark every prior explicit interface row as Deleting and
+		//     return an empty list. Reads after this should reflect the
+		//     auto contract (no explicit interfaces) rather than the
+		//     stale rows that pre-dated the mode switch.
+		//   - Explicit interfaces in the request: create the new rows
+		//     and mark the previous rows as Deleting (existing behavior).
+		//   - Neither (no interface change, not switching to auto):
+		//     carry the existing rows forward.
+		switch {
+		case ui.NetworkAuto && len(apiRequest.Interfaces) == 0:
```
thossain-nv, line 3016: Are we supporting switching a regular Instance with Interfaces to auto interface?

api/pkg/api/model/instance.go
```diff
@@ -368,8 +368,15 @@ type APIInstanceCreateRequest struct {
...
-	// Interfaces is the list of Interfaces to create for the Instance
+	// Interfaces is the list of Interfaces to create for the Instance.
+	// Mutually exclusive with `Auto`: when `Auto` is true this MUST be empty.
 	Interfaces []APIInterfaceCreateOrUpdateRequest `json:"interfaces"`
+	// Auto, when true, asks NICo to auto-resolve the Instance's network
+	// interfaces from the host's HostInband network segments. Intended for
+	// instances on zero-DPU hosts (or hosts with their DPU in NIC mode).
+	// When true, `Interfaces` MUST be empty. The resolved per-interface
+	// details surface in the Instance's status.
+	Auto bool `json:"auto"`
```
thossain-nv, line 379: `Auto` doesn't seem descriptive enough for an Instance level attribute. Shouldn't it be `NetworkAuto` or better yet `AutoNetwork`?

api/pkg/api/model/instance.go
```diff
@@ -421,8 +428,14 @@ type APIBatchInstanceCreateRequest struct {
...
 	UserData *string `json:"userData"`
-	// Interfaces is the list of Interfaces to create for each instance (shared across all instances)
+	// Interfaces is the list of Interfaces to create for each instance (shared across all instances).
+	// Mutually exclusive with `Auto`: when `Auto` is true this MUST be empty.
 	Interfaces []APIInterfaceCreateOrUpdateRequest `json:"interfaces"`
+	// Auto, when true, asks NICo to auto-resolve each Instance's network
+	// interfaces from the host's HostInband network segments. Intended for
+	// instances on zero-DPU hosts (or hosts with their DPU in NIC mode).
+	// When true, `Interfaces` MUST be empty.
+	Auto bool `json:"auto"`
```
thossain-nv, line 438: Same as above.

api/pkg/api/model/instance.go
```diff
@@ -1071,8 +1108,14 @@ type APIInstanceUpdateRequest struct {
...
 	SecondaryVpcIDs []string `json:"secondaryVpcIds"`
-	// Interfaces is the list of Interfaces to update for the Instance
+	// Interfaces is the list of Interfaces to update for the Instance.
+	// Mutually exclusive with `Auto`: when `Auto` is true this MUST be empty.
 	Interfaces []APIInterfaceCreateOrUpdateRequest `json:"interfaces"`
+	// Auto, when set, asks NICo to auto-resolve the Instance's network
+	// interfaces from the host's HostInband network segments. `nil` leaves
+	// the value unchanged; `true` re-resolves; `false` returns to explicit
+	// interface configuration. When `true`, `Interfaces` MUST be empty.
+	Auto *bool `json:"auto"`
```
thossain-nv, line 1118: Same as above.

api/pkg/api/model/instance.go
```diff
@@ -1587,6 +1644,11 @@ type APIInstance struct {
 	TpmEkCertificate *string `json:"tpmEkCertificate"`
 	// Status is the status of the Instance
 	Status string `json:"status"`
+	// Auto is true when this Instance had its network interfaces
+	// auto-resolved by NICo from the host's HostInband segments. When
+	// true, `Interfaces` reflects the resolved set; the caller's request
+	// list was empty.
+	Auto bool `json:"auto"`
```
thossain-nv, line 1651: Same as above.

db/pkg/db/model/instance_test.go
```diff
@@ -1584,6 +1584,7 @@ type InstanceWithAllocation struct {
 	AlwaysBootWithCustomIpxe               bool                                    `bun:"always_boot_with_custom_ipxe,notnull"`
 	PhoneHomeEnabled                       bool                                    `bun:"phone_home_enabled,notnull"`
 	UserData                               *string                                 `bun:"user_data"`
+	NetworkAuto                            bool                                    `bun:"network_auto,notnull"`
```
thossain-nv, line 1587: Could this be `AutoNetwork` instead which reads to the user/dev as _automatic network_?

thossain-nv (commented):
@chet Sorry I missed reviewing this earlier this week. Please take a look at the suggestions when convenient.

PR #576, opened 2026-05-27
feat: Add --api-name flag and OIDC username/password env vars

kfelternv:
Follow up PR to make the api name a flag so that it shows up in the help command in the cli


thossain-nv (approved):
Thanks @kfelternv

PR #580, opened 2026-05-27
chore: Switch file headers to two-line SPDX license style

kfelternv:
This updates all files to use the 2 line header for licenses instead of the larger license header. This helps agentic development by reducing code per file. This also updates the scripts in the makefile that add the headers to use the smaller version of the header.


thossain-nv (approved):
Looks good, thanks Kyle.

PR #589, opened 2026-05-29
fix: Read service-account flag from request context not issuer config

parmani-nv:
## Description
`GetCurrentServiceAccountHandler` decided whether the caller was a service
account by scanning every configured issuer and returning `true` if any
issuer had service accounts enabled. This misclassified ordinary user
tokens as service accounts whenever the deployment had any
service-account-capable issuer configured.
This PR makes the handler read the per-request flag that the auth
processors already record via `cauth.SetIsServiceAccountInContext`
(`GetIsServiceAccountFromContext`), so `/service-account/current` reflects
the actual authenticated caller rather than global config. Keycloak and
custom (`claimMapping.isServiceAccount`) tokens both funnel into the same
context flag, so this works uniformly across origins.
It also restricts `claimMapping.isServiceAccount` to disconnected mode,
matching the existing issuer-level `serviceAccount` guard, since it grants
admin roles that must not be self-asserted from local config when a cloud
identity provider is the source of truth.

## Type of Change

## Services Affected

## Related Issues (Optional)

None

## Breaking Changes

## Testing

### DIFF
--- a/api/internal/config/config.go
+++ b/api/internal/config/config.go
@@ -621,6 +621,9 @@ func (c *Config) ValidateIssuersConfig(issuers []IssuerConfig) error {
 				if mapping.OrgAttribute != "" {
 					return fmt.Errorf("issuer %s: claimMapping %d: orgAttribute cannot be specified when isServiceAccount is true", issuer.Name, j)
 				}
+				if !c.GetEnvDisconnected() {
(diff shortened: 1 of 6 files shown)


thossain-nv (approved):
Looks good, thanks @parmani-nv

